PR #17, opened 2017-06-29
Fixes to make Travis Green

kevmoo:


kevmoo:
> Any idea how we can get analyzer coverage on the examples, too? They need their own pubspec's, which I guess is where this problem comes from.

Sadly, no.

Not as long as we're using the `dart_task` thing.

We'd have to go back to using `travis.sh` so we can run tasks within a sub directory

kevmoo:
...or create a test in `root/test` that runs the analyzer process on the subdirectory – a bit of a hack, but it'd work

jakobr-google:
I think I'll just add a custom step to analyze the examples.

Issue #12, opened 2017-06-21
Add server options, channel options, per-RPC options


jakobr-google:
Like TLS settings, authentication, etc.

Issue #8, opened 2017-06-21
Implement timeout/deadline and cancelation


jakobr-google:

jakobr-google:
Blocked on #13.

jakobr-google:
Cancellation is in. Timeout header is sent to server, but client and server don't actually handle timeouts yet.

PR #22, opened 2017-07-07
Custom metadata and call cancellation.

jakobr-google:
Added an example showing how to send/receive custom metadata, and handle
call cancellation.

Implemented the remaining parts of metadata and cancel handling.

Addresses part of #8 and #12.

### DIFF
--- a/example/metadata/bin/client.dart
+++ b/example/metadata/bin/client.dart
@@ -4,6 +4,6 @@
 
 import 'package:metadata/src/client.dart';
 
-main(List<String> args) async {
-  await new Client().main(args);
+main(List<String> args) {
+  new Client().main(args);
 }
--- a/example/metadata/bin/client.dart
+++ b/example/metadata/bin/client.dart
@@ -1,0 +1,9 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:metadata/src/client.dart';
+
+main(List<String> args) async {
+  await new Client().main(args);
+}
--- a/example/metadata/bin/server.dart
+++ b/example/metadata/bin/server.dart
@@ -4,6 +4,6 @@
 
 import 'package:metadata/src/server.dart';
 
-main(List<String> args) async {
-  await new Server().main(args);
+main(List<String> args) {
+  new Server().main(args);
 }
--- a/example/metadata/bin/server.dart
+++ b/example/metadata/bin/server.dart
@@ -1,0 +1,9 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:metadata/src/server.dart';
+
+main(List<String> args) async {
+  await new Server().main(args);
+}
--- a/example/metadata/lib/src/client.dart
+++ b/example/metadata/lib/src/client.dart
@@ -12,12 +12,9 @@ class Client {
   ClientChannel channel;
   MetadataClient stub;
 
-  Client() {
+  Future<Null> main(List<String> args) async {
     channel = new ClientChannel('127.0.0.1', port: 8080);
     stub = new MetadataClient(channel);
-  }
-
-  Future<Null> main(List<String> args) async {
     // Run all of the demos in order.
     await runEcho();
     await runEchoDelayCancel();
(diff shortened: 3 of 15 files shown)


example/metadata/bin/client.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:metadata/src/client.dart';
+
+main(List<String> args) async {
+  await new Client().main(args);
```
kasperl: This looks nice, but you could also just mark this function non-async (drop the `async` modifier) and return the result of calling `new Client().main(args)`, right?

example/metadata/bin/server.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:metadata/src/server.dart';
+
+main(List<String> args) async {
+  await new Server().main(args);
```
kasperl: This looks nice, but you could also just mark this function non-async (drop the async modifier) and return the result of calling new Server().main(args), right?

example/metadata/lib/src/client.dart
```diff
@@ -0,0 +1,110 @@
...
+import 'package:grpc/grpc.dart';
+
+import 'generated/metadata.pbgrpc.dart';
+
+class Client {
+  ClientChannel channel;
+  MetadataClient stub;
+
+  Client() {
+    channel = new ClientChannel('127.0.0.1', port: 8080);
```
kasperl: It would feel a bit more obvious to initialize these in `main` where `channel` is also closed.

example/metadata/lib/src/generated/metadata.pb.dart
```diff
@@ -0,0 +1,109 @@
+///
```
kasperl: It looks a little bit funky with the two lines of empty doc comments in the header.

example/metadata/bin/client.dart
jakobr-google: Yep, looks like I can!

lib/src/client.dart
```diff
@@ -76,32 +104,44 @@ class ClientCall<Q, R> implements Response {
...
+    final timeout = options?.timeout ?? _channel.options?.timeout;
     // TODO(jakobr): Populate HTTP-specific headers in connection?
     final headers = <Header>[
       _methodPost,
       _schemeHttp,
       new Header.ascii(':path', _method.path),
       new Header.ascii(':authority', _channel.host),
+    ];
+    if (timeout != null) {
+      headers.add(new Header.ascii('grpc-timeout', '${timeout.inSeconds}S'));
```
kasperl: Are you sure you are happy with the rounding implied by `inSeconds`?

example/metadata/lib/src/generated/metadata.pb.dart
```diff
@@ -0,0 +1,109 @@
...
+  Empty() : super();
+  Empty.fromBuffer(List<int> i, [ExtensionRegistry r = ExtensionRegistry.EMPTY])
+      : super.fromBuffer(i, r);
+  Empty.fromJson(String i, [ExtensionRegistry r = ExtensionRegistry.EMPTY])
+      : super.fromJson(i, r);
+  Empty clone() => new Empty()..mergeFromMessage(this);
+  BuilderInfo get info_ => _i;
+  static Empty create() => new Empty();
+  static PbList<Empty> createRepeated() => new PbList<Empty>();
+  static Empty getDefault() {
```
kasperl: This could be simplified by just lazily initializing `_defaultInstance` to `new _ReadonlyEmpty`. Would it ever be possible to have the `_ReadOnlyEmpty` class be a compile-time constant?

example/metadata/lib/src/generated/metadata.pb.dart
jakobr-google: Very funky. Not sure why it was done this way, or why the 'generated code' comment isn't a dartdoc comment...

There's an open issue on including the comments from the .proto file in the generated .dart files, so I'm planning to address it at that point. Comments were originally left out because they caused larger load times in Dartium, I believe.

example/metadata/lib/src/generated/metadata.pb.dart
jakobr-google: I'm hoping to get rid of the _ReadOnly classes completely, as part of the 'immutability' effort.

lib/src/client.dart
jakobr-google: I'm not, but my brain's stopped working for today. You'll notice that I'm also adding an extra timeout of 5S just below this... Fix coming up.

Issue #10, opened 2017-06-21
Implement gRPC interoperability tests


jakobr-google:
See https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md

jakobr-google:
Client-side interop tests landed in #32. The compression tests are currently missing, since we don't support compression yet.

jakobr-google:
Compression is tracked in #6, so closing this one.

PR #25, opened 2017-07-11
Basic server interoperability tests.

jakobr-google:
Implements the basic server interoperability tests for gRPC compliance.
Compressed tests, TLS, and authentication are not implemented yet, but
basic test cases pass against the C++ gRPC interoperability client.

Part of #10.

### DIFF
--- a/interop/bin/server.dart
+++ b/interop/bin/server.dart
@@ -1,0 +1,113 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:args/args.dart';
+import 'package:grpc/grpc.dart';
+
+import 'package:interop/src/generated/empty.pb.dart';
+import 'package:interop/src/generated/messages.pb.dart';
+import 'package:interop/src/generated/test.pbgrpc.dart';
+
+const _headerEchoKey = 'x-grpc-test-echo-initial';
+const _trailerEchoKey = 'x-grpc-test-echo-trailing-bin';
+
+class TestService extends TestServiceBase {
+  @override
+  void $onMetadata(ServiceCall context) {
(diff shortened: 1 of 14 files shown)


interop/protos/messages.proto
```diff
@@ -0,0 +1,184 @@
...
+  // stream.
+  PayloadType response_type = 1;
+
+  // Configuration for each expected response message.
+  repeated ResponseParameters response_parameters = 2;
+
+  // Optional input payload sent along with the request.
+  Payload payload = 3;
+
+  // Whether server should return a given status
```
johnniwinther: `status` -> `status.`

interop/bin/server.dart
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+Future<Null> main(List<String> args) async {
+  final argumentParser = new ArgParser();
+  argumentParser.addOption('port', defaultsTo: '8080');
+  argumentParser.addOption('use_tls', defaultsTo: 'false');
+  final arguments = argumentParser.parse(args);
+  final port = int.parse(arguments['port']);
+  //final useTls = arguments['use_tls'] == 'true';
```
johnniwinther: Remove or document why its commented out.

lib/src/server.dart
```diff
@@ -284,13 +302,15 @@ class ServerHandler {
       final bytes = _descriptor.responseSerializer(response);
       _stream.sendData(GrpcHttpEncoder.frame(bytes));
     } catch (error) {
```
johnniwinther: Shouldn't we `_sinkIncoming` for this error case, like above?

interop/bin/server.dart
jakobr-google: Removed. I'll re-add it when I actually implements TLS handling.

interop/protos/messages.proto
jakobr-google: Done.

lib/src/server.dart
jakobr-google: Probably, yes. Done.

PR #26, opened 2017-07-12
Add basic client tests.

jakobr-google:


test/src/client_utils.dart
```diff
@@ -0,0 +1,181 @@
...
+class MockChannel extends Mock implements ClientChannel {}
+
+typedef ServerMessageHandler = void Function(StreamMessage message);
+
+class TestClient {
+  final ClientChannel _channel;
+
+  static final _$unary = new ClientMethod<int, int>(
+      '/Test/Unary',
+      (int value) => new List.filled(value, 0),
```
johnniwinther: Extract these as `mockEncode` and `mockDecode` functions to use in the client methods and in the request handlers in the tests.

test/client_test.dart
```diff
@@ -0,0 +1,314 @@
...
+  tearDown(() {
+    harness.tearDown();
+  });
+
+  test('Unary calls work end-to-end', () async {
+    void _handleRequest(StreamMessage message) {
+      expect(message, new isInstanceOf<DataStreamMessage>());
+      expect(message.endStream, false);
+      final data = new GrpcHttpDecoder().convert(message) as GrpcData;
+      expect(data.data.length, 17);
```
johnniwinther: Extract these numbers to clearly show that they magic values.

jakobr-google: Done.

test/src/client_utils.dart
jakobr-google: Done.

PR #29, opened 2017-07-20
Add TLS support.

### LINKED ISSUE #4: Support secure sockets
Initial implementation only supports TLS (HTTPS) sockets. We should support plain sockets, too.

jakobr-google:
Added ChannelOptions, which are used to specify options on a
ClientChannel. At the moment, only TLS options are supported.

Moved CallOptions from ClientChannel to a new Client stub base class.
Per-RPC call options are now specified on the stub instead of on the
channel, allowing several stubs with different options to share the same
channel.

Added support for TLS on the server side. TLS options are specified when
creating the Server.

Fixes #4.


lib/src/server.dart
```diff
@@ -56,25 +56,54 @@ abstract class Service {
...
-  void addService(Service service) {
-    _services[service.$name] = service;
+  factory Server.secure(List<Service> services,
+      {List<int> certificate,
+      String certificatePassword,
+      List<int> privateKey,
+      String privateKeyPassword,
+      int port}) {
+    final context = createSecurityContext(true);
+    if (privateKey != null) {
```
mravn-google: Again: what happens if passwords are null, but certificate and/or private key are not? Do we fail in the right way, if encryption is used?

lib/src/server.dart
```diff
@@ -56,25 +56,54 @@ abstract class Service {
 /// Listens for incoming gRPC calls, dispatching them to a [ServerHandler].
 class Server {
   final Map<String, Service> _services = {};
-  ServerSocket _server;
+  final int port;
+  final SecurityContext _securityContext;
 
+  ServerSocket _insecureServer;
+  SecureServerSocket _secureServer;
```
mravn-google: Not entirely clear why we should need both. Don't the two types have a common ancestor with a `close` method?

lib/src/client.dart
```diff
@@ -80,6 +97,49 @@ class ClientMethod<Q, R> {
   ClientMethod(this.path, this.requestSerializer, this.responseDeserializer);
 }
 
+/// Runtime options for a RPC call.
```
jakobr-google: I went with 'an RPC.' :)

lib/src/client.dart
```diff
@@ -80,6 +97,49 @@ class ClientMethod<Q, R> {
...
+        sanitizedMetadata[lowerCaseKey] = value;
+      }
+    });
+    return new CallOptions._(new Map.unmodifiable(sanitizedMetadata), timeout);
+  }
+
+  CallOptions mergedWith(CallOptions other) {
+    if (other == null) return this;
+    final mergedMetadata = new Map.from(metadata)..addAll(other.metadata);
+    final mergedTimeout = other.timeout ?? timeout;
```
jakobr-google: It's a deliberate choice, yes. The `Client` has a default `CallOptions` that might specify a default timeout, which you can override to something specific per RPC.

In general, anything specified in `other` will override whatever was specified in `this`.

lib/src/server.dart
```diff
@@ -56,25 +56,54 @@ abstract class Service {
 /// Listens for incoming gRPC calls, dispatching them to a [ServerHandler].
```
jakobr-google: Done.

lib/src/server.dart
jakobr-google: You'd think so, but no.

`ServerSocket` implements `Stream<Socket>`, and `SecureServerSocket` extends `Stream<SecureSocket>`, but `Stream` doesn't define a `close` method. They each define their own.

lib/src/server.dart
jakobr-google: Yep. I added a comment.

PR #31, opened 2017-09-20
Preparation for RPC multiplexing

jakobr-google:
First stage of separating Connection from Channel. A Channel manages
multiple Connections, and chooses which Connection to send an RPC on.

In this change, the Channel still creates a Connection for each RPC.
Managing the Connection life-cycle comes in a later change.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -26,9 +26,10 @@ class ChannelOptions {
   final bool _useTls;
   final List<int> _certificateBytes;
   final String _certificatePassword;
+  final String authority;
 
   const ChannelOptions._(this._useTls,
-      [this._certificateBytes, this._certificatePassword]);
+      [this._certificateBytes, this._certificatePassword, this.authority]);
 
   /// Enable TLS using the default trust store.
   const ChannelOptions() : this._(true);
@@ -37,8 +38,9 @@ class ChannelOptions {
   const ChannelOptions.insecure() : this._(false);
 
   /// Enable TLS and specify the [certificate]s to trust.
-  ChannelOptions.secure({List<int> certificate, String password})
-      : this._(true, certificate, password);
+  ChannelOptions.secure(
+      {List<int> certificate, String password, String authority})
+      : this._(true, certificate, password, authority);
 
   SecurityContext get securityContext {
     if (!_useTls) return null;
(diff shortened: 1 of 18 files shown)


lib/src/client.dart
```diff
@@ -51,42 +53,154 @@ class ChannelOptions {
...
+      _grpcAcceptEncoding,
+      _userAgent,
+    ]);
+    options.metadata.forEach((key, value) {
+      headers.add(new Header.ascii(key, value));
+    });
+    return headers;
+  }
+
+  /// Shut down this connection.
```
mravn-google: [Effective dart](https://www.dartlang.org/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs) would have you say "Shuts down this connection".

Applies several times below.

lib/src/client.dart
```diff
@@ -51,42 +53,154 @@ class ChannelOptions {
...
+  ///
+  /// All open calls are terminated immediately, and no further calls may be
+  /// made on this connection.
+  Future<Null> terminate() {
+    _isShutdown = true;
+    // TODO(jakobr): Manage streams, close them immediately.
+    return _transport.terminate();
+  }
+
+  ClientTransportStream sendRequest(
```
mravn-google: Dartdoc?

lib/src/client.dart
```diff
@@ -51,42 +53,154 @@ class ChannelOptions {
...
+  static final _contentTypeGrpc =
+      new Header.ascii('content-type', 'application/grpc');
+  static final _teTrailers = new Header.ascii('te', 'trailers');
+  static final _grpcAcceptEncoding =
+      new Header.ascii('grpc-accept-encoding', 'identity');
+  static final _userAgent = new Header.ascii('user-agent', 'dart-grpc/0.2.0');
+
+  final ClientTransportConnection _transport;
+
+  bool _isShutdown = false;
```
mravn-google: Could this state be avoided by using the channel's `_isShutdown` field instead?

lib/src/client.dart
```diff
@@ -51,42 +53,154 @@ class ChannelOptions {
...
+      headers.add(new Header.ascii(key, value));
+    });
+    return headers;
+  }
+
+  /// Shut down this connection.
+  ///
+  /// No further calls may be made on this connection, but existing calls
+  /// are allowed to finish.
+  Future<Null> shutdown() {
```
mravn-google: Client code might expect this operation to be idempotent. Is it? Dartdoc should state what happens if you call it twice.

Same for `terminate` and the two similar methods on `ClientChannel`.

lib/src/client.dart
```diff
@@ -51,42 +53,154 @@ class ChannelOptions {
...
+    _shutdownCheck(() => socket.destroy());
+    if (securityContext != null) {
+      socket = await SecureSocket.secure(socket,
+          host: authority, context: securityContext);
+      _shutdownCheck(() => socket.destroy());
     }
-    _sockets.add(socket);
-    final connection = new ClientTransportConnection.viaSocket(socket);
+    final connection =
+        new ClientConnection(new ClientTransportConnection.viaSocket(socket));
```
mravn-google: Seems a bit weird that a `ClientConnection` is created from a `ClientTransportConnection` rather than a `ClientTransport`.

lib/src/client.dart
```diff
@@ -51,42 +53,154 @@ class ChannelOptions {
...
     return connection;
   }
 
-  /// Close all connections made on this [ClientChannel].
-  Future<Null> close() async {
-    await Future.wait(_connections.map((c) => c.finish()));
-    _connections.clear();
-    await Future.wait(_sockets.map((s) => s.close()));
-    _sockets.clear();
+  ClientCall<Q, R> createCall<Q, R>(
```
mravn-google: Dartdoc?

lib/src/client.dart
```diff
@@ -137,103 +251,75 @@ class Client {
...
-  final Completer<Map<String, String>> _headers = new Completer();
-  final Completer<Map<String, String>> _trailers = new Completer();
+  final _headers = new Completer<Map<String, String>>();
+  final _trailers = new Completer<Map<String, String>>();
   bool _hasReceivedResponses = false;
 
   Map<String, String> _headerMetadata;
 
   TransportStream _stream;
+  // ignore: close_sinks
```
mravn-google: This comment could be more self-explanatory to help future maintainers.

lib/src/client.dart
```diff
@@ -359,6 +446,7 @@ class ClientCall<Q, R> implements Response {
     _responses.addError(error);
     _timeoutTimer?.cancel();
     _responses.close();
+    _requestSubscription.cancel();
     _responseSubscription?.cancel();
     _stream.terminate();
```
mravn-google: `_stream` may be null here I think.

lib/src/client.dart
jakobr-google: True. This'll be idempotent when I implement the connection states, but isn't currently.

PR #32, opened 2017-09-25
Added client-side interop tests.

jakobr-google:


interop/lib/src/client.dart
```diff
@@ -0,0 +1,1027 @@
...
+  /// * metadata with key `"x-grpc-test-echo-initial"` and value
+  ///   `"test_initial_metadata_value"`is received in the initial metadata for
+  ///   calls in Procedure steps 1 and 2.
+  /// * metadata with key `"x-grpc-test-echo-trailing-bin"` and value `0xababab`
+  ///   is received in the trailing metadata for calls in Procedure steps 1 and
+  ///   2.
+  Future<Null> customMetadata() async {
+    void validate(Map<String, String> headers, Map<String, String> trailers) {
+      if (headers[_headerEchoKey] != _headerEchoData) {
+        throw 'Invalid header data recived.';
```
mravn-google: `received`

interop/lib/src/client.dart
```diff
@@ -0,0 +1,1027 @@
...
+      }
+      if (trailers[_trailerEchoKey] != _trailerEchoData) {
```
mravn-google: [puzzled] The trailer data is not really binary, but a hex string. How is that any different from the ascii header data?

interop/lib/src/client.dart
```diff
@@ -0,0 +1,1027 @@
...
+  /// Client asserts:
+  /// * Call completed with status CANCELLED
+  Future<Null> cancelAfterFirstResponse() async {
+    final requests = new StreamController<StreamingOutputCallRequest>();
+    final call = client.fullDuplexCall(requests.stream);
+    final completer = new Completer();
+
+    var receivedResponse = false;
+    call.listen((response) {
+      if (receivedResponse) throw 'Received too many responses.';
```
mravn-google: Who receives the exception thrown here? Should this not complete `completer` with an error instead? Same on line 969.

interop/lib/src/client.dart
```diff
@@ -0,0 +1,1027 @@
...
+    try {
+      await for (final _ in call) {
+        throw 'Unexpected response received.';
+      }
+      throw 'Expected exception.';
+    } on GrpcError catch (e) {
+      if (e.code != StatusCode.deadlineExceeded) {
+        throw 'Unexpected status code ${e.code} - ${e.message}.';
+      }
+      return;
```
mravn-google: A `return` statement here changes nothing, right?

interop/lib/src/client.dart
jakobr-google: It isn't really, but some languages have APIs to set/get binary values, which automatically append/strip the '-bin' suffix and en/decode the binary data as base64.

And I just forgot to base64 encode the hex value in the test description. Fixed. It's still just a string, though, which is why the test passed...

interop/lib/src/client.dart
jakobr-google: No-one can receive this exception, so yeah, it's probably better to complete with an error. (End result is the same, since no-one catches the error thrown by `await completer.future`, but at least someone *could*.)

interop/lib/src/client.dart
jakobr-google: Right. Removed.

Issue #5, opened 2017-06-21
Implement connection management


jakobr-google:
RPC calls to the same endpoint should share a single connection (or pool of connections if the service is load balanced). Implementation should respect connection timeouts.

See https://github.com/dart-lang/appengine/blob/master/lib/src/grpc_api_impl/grpc.dart#L293 for reference.

jakobr-google:
RPCs on the same Channel now share a single Connection. What's missing here is support for load balancing RPCs across several Connections, and support for naming schemes other than DNS.

galihputera:
Hi guys. Are we still working on this issue?

PR #34, opened 2017-10-11
Multiplex RPCs on a single connection.

jakobr-google:
A Channel will now multiplex RPCs on a single managed connection. The
connection will attempt to reconnect on failure, and will close the
underlying transport connection if no RPCs have been made for a while.

Part of #5.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -7,7 +7,9 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'dart:math';
 import 'package:http2/transport.dart';
+import 'package:meta/meta.dart';
 
 import 'shared.dart';
 import 'status.dart';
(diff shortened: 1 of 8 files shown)


lib/src/client.dart
```diff
@@ -95,34 +168,166 @@ class ClientConnection {
...
+  void _setState(ConnectionState state) {
+    _state = state;
+    if (onStateChanged != null) {
+      onStateChanged(this);
+    }
+  }
+
+  void _handleIdleTimeout() {
+    if (_timer == null) return;
+    _timer = null;
```
mravn-google: Use `_cancelTimer` instead of setting `_timer` to `null` in more than one place. Conveys intent better, I think. Plus it cancels unwanted timer events instead of attempting to ignore them when they occur. Unwanted timer events can make debugging hard, but I think your attempt at ignoring them may even fail, if a new timer has been set up in the mean time? You may need to receive the timer object with timer events and compare to current `_timer` before reacting.

lib/src/client.dart
```diff
@@ -95,34 +168,166 @@ class ClientConnection {
...
+
+  void _handleTransientFailure(error) {
+    _transport = null;
+    if (_state == ConnectionState.shutdown || _state == ConnectionState.idle) {
+      return;
+    }
+    // TODO(jakobr): Log error.
+    _cancelTimer();
+    if (!_hasPendingCalls()) {
+      // If there are no pending calls, just go directly to idle.
```
mravn-google: [nit] This comment just repeats what the code already says.

test/client_test.dart
```diff
@@ -310,4 +300,82 @@ void main() {
...
+      connectionStates.add(connection.state);
+    };
+
+    harness.channelOptions.idleTimeout = const Duration(microseconds: 10);
+
+    await makeUnaryCall();
+    harness.signalIdle();
+    expect(
+        connectionStates, [ConnectionState.connecting, ConnectionState.ready]);
+    await new Future.delayed(const Duration(microseconds: 100));
```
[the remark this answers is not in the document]
jakobr-google: Done. I kind of wanted to make sure I didn't get another state change after idle, but I don't really think it matters.

lib/src/client.dart
```diff
@@ -21,38 +23,95 @@ const _reservedHeaders = const [
...
 
   /// Disable TLS. RPCs are sent in clear text.
-  const ChannelOptions.insecure() : this._(false);
-
-  /// Enable TLS and specify the [certificate]s to trust.
-  ChannelOptions.secure(
-      {List<int> certificate, String password, String authority})
-      : this._(true, certificate, password, authority);
+  const ChannelOptions.insecure(
+      {Duration idleTimeout,
```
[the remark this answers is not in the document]
jakobr-google: Hard to believe, but yes. It insists. (And the Travis checks enforce using `dartfmt`.)

lib/src/client.dart
```diff
@@ -21,38 +23,95 @@ const _reservedHeaders = const [
...
+    }
+    final context = SecurityContext.defaultContext;
+    if (SecurityContext.alpnSupported) {
+      context.setAlpnProtocols(supportedAlpnProtocols, false);
     }
     return context;
   }
 }
 
+enum ConnectionState {
```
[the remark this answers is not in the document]
jakobr-google: I agree, but these names are defined by gRPC: https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md

lib/src/client.dart
jakobr-google: True, it does now. Removed.

lib/src/client.dart
jakobr-google: I've added in a check for the current state. Idle timeout should only fire while in the `ready` state. We only set the timeout when there are no more active streams, and we explicitly clear it when the connection is active again, so if we get a timeout callback, and:

* _state == ready
* _timer != null

then at least there are no active streams. Worst case, we end up closing an idle connection too soon, but the following would have to happen:

1. Connection has been idle long enough for the timer to fire and schedule an event to call `_handleTimeout()`.
1. Before `_handleTimeout()` is actually called, the currently running code makes an RPC, making the connection active (and clearing `_timer` in the process).
1. Still before the scheduled `_handleTimeout()` call is invoked, the RPC manages to finish, and a new idle timeout is scheduled.
1. The original `_handleTimeout()` call finally runs, and closes the connection.

I don't think it's possible to complete an RPC without processing other events in the queue (especially since the `onActiveStateChange` call to mark the transport idle is delivered in a Future event, which would come later in the queue), so I we should be good now.

PR #39, opened 2017-12-07
Added support for metadata providers.

jakobr-google:
Provide a hook for metadata providers that need to generate their
metadata for each RPC. An example is authorization, where the provider
may need to refresh a token before it can provide the header.

Add stackdriver logging examplei to demonstrate calling a Google API.

Updated other examples to protobuf 0.6.0 (protoc plugin 0.7.8).

Review note: All the \*.pb\*.dart files (example/googleapis/lib/src/generated/*, etc.) are generated from .proto source. Unless you're feeling really diligent, there's no need to review them.

### DIFF
--- a/lib/src/client.dart
+++ b/lib/src/client.dart
@@ -87,7 +87,7 @@ class ChannelOptions {
         ..setTrustedCertificatesBytes(_certificateBytes,
             password: _certificatePassword);
     }
-    final context = SecurityContext.defaultContext;
+    final context = new SecurityContext(withTrustedRoots: true);
     if (SecurityContext.alpnSupported) {
       context.setAlpnProtocols(supportedAlpnProtocols, false);
     }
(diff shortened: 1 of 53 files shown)


lib/src/client.dart
```diff
@@ -396,31 +399,40 @@ class ClientMethod<Q, R> {
   ClientMethod(this.path, this.requestSerializer, this.responseDeserializer);
 }
 
+/// Provides per-RPC metadata.
+///
+/// Metadata providers will be invoked for every RPC, and can add their own
+/// metadata to the RPC. If the function returns a [Future], the RPC will await
+/// completion of the returned [Future] before transmitting the request.
+typedef FutureOr<Null> MetadataProvider(Map<String, String> metadata);
```
mravn-google: Please document that the `MetadataProvider` is supposed to mutate the provided `metadata` map before it returns or before its returned `Future` completes. And at the risk of stating the obvious: a `MetadataProvider` may assume that the provided map is not null.

lib/src/client.dart
```diff
@@ -396,31 +399,40 @@ class ClientMethod<Q, R> {
...
-      if (!lowerCaseKey.startsWith(':') &&
-          !_reservedHeaders.contains(lowerCaseKey)) {
-        sanitizedMetadata[lowerCaseKey] = value;
-      }
-    });
-    return new CallOptions._(new Map.unmodifiable(sanitizedMetadata), timeout);
+  factory CallOptions(
+      {Map<String, String> metadata,
+      Duration timeout,
+      List<MetadataProvider> providers}) {
```
mravn-google: You may want to document here that the `MetadataProviders` are invoked with the given `metadata` map in turn and may each modify it to their liking. Last `MetadataProvider` decides the outcome (modulo sanitization).

PR #41, opened 2017-12-14
Reorganize code for 0.2.0 release.

jakobr-google:
Mostly just moving code around. Actual changes:

* ServerHandler now implements ServiceCall directly, instead of having ServiceCall be a proxy for ServerHandler.
* Updated pubspec.yaml and CHANGELOG for 0.2.0 release.


lib/src/client/call.dart
```diff
@@ -0,0 +1,285 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:http2/transport.dart';
+
+import '../shared/status.dart';
+import '../shared/streams.dart';
+
+import 'client.dart';
```
mravn-google: This dependency is part of a cycle. You may be able to break it by moving `ClientMethod` here or to its own `client/method.dart` file.

lib/src/client/channel.dart
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import '../shared/status.dart';
+
+import 'call.dart';
+import 'client.dart';
```
mravn-google: Same here.

PR #55, opened 2018-02-02
Authenticated cloud-to-prod interop tests.

### LINKED ISSUE #53: Implement cloud-to-prod-auth interop tests
As described in https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md.

We're missing these tests:
* compute_engine_creds
* jwt_token_creds
* oauth2_auth_token
* per_rpc_creds

(other language's interop clients also have a service_account_creds test, but I don't know if that one is required)

package:googleapis_auth has support for these, so they just need to be wired up.

jakobr-google:
Added authentication provider classes, and wired up the auth interop
tests.

Refactored connection logic to throw initial connection errors early.

Fixes #53

### DIFF
--- a/lib/src/auth/auth.dart
+++ b/lib/src/auth/auth.dart
@@ -1,0 +1,164 @@
+// Copyright (c) 2017, the gRPC project authors. Please see the AUTHORS file
+// for details. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: grpc
 description: Dart implementation of gRPC.
-version: 0.2.1
+version: 0.3.0
```
mravn-google: Missing changelog entry to go with this version bump.

lib/src/auth/auth.dart
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2017, the gRPC project authors. Please see the AUTHORS file
```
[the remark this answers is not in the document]
jakobr-google: I think you're right. I seem to recall celebrating New Year's about a month ago.

pubspec.yaml
jakobr-google: And I was doing so well until now. Coming up!

PR #60, opened 2018-02-26
Split out TLS credentials to a separate class.

jakobr-google:
Add a 'bad certificate handler' to the new ChannelCredentials, which can
be used to override certificate validation (for example, to allow
auto-generated self-signed certificates during development).

Also fixed a bug in Server.shutdown().

### DIFF
--- a/lib/src/client/options.dart
+++ b/lib/src/client/options.dart
@@ -91,9 +91,9 @@ class ChannelCredentials {
 
 /// Options controlling how connections are made on a [ClientChannel].
 class ChannelOptions {
+  final ChannelCredentials credentials;
   final Duration idleTimeout;
   final BackoffStrategy backoffStrategy;
-  final ChannelCredentials credentials;
 
   const ChannelOptions(
       {ChannelCredentials credentials,
--- a/lib/src/client/options.dart
+++ b/lib/src/client/options.dart
@@ -39,43 +39,40 @@ Duration defaultBackoffStrategy(Duration lastBackoff) {
   return nextBackoff < _maxBackoff ? nextBackoff : _maxBackoff;
 }
 
-/// Options controlling how connections are made on a [ClientChannel].
-class ChannelOptions {
+/// Handler for checking certificates that fail validation. If this handler
+/// returns `true`, the bad certificate is allowed, and the TLS handshake can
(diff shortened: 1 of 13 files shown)


lib/src/client/options.dart
```diff
@@ -92,6 +89,22 @@ class ChannelOptions {
...
+  final ChannelCredentials credentials;
+
+  const ChannelOptions(
+      {ChannelCredentials credentials,
+      Duration idleTimeout,
+      BackoffStrategy backoffStrategy =
+          defaultBackoffStrategy}) // Remove when dart-lang/sdk#31066 is fixed.
+      : this.credentials = credentials ?? const ChannelCredentials.secure(),
+        this.idleTimeout = idleTimeout ?? defaultIdleTimeout,
+        this.backoffStrategy = backoffStrategy ?? defaultBackoffStrategy;
```
mravn-google: [nit] Make ordering of fields consistent with that of constructor parameters.
[another nit] My Flutter eyes are hurting by the erratic indentation produced by `dartfmt` here.

lib/src/server/server.dart
```diff
@@ -100,14 +100,14 @@ class Server {
...
+  Future<Null> shutdown() async {
     final done = _connections.map((connection) => connection.finish()).toList();
     if (_insecureServer != null) {
       done.add(_insecureServer.close());
     }
     if (_secureServer != null) {
       done.add(_secureServer.close());
     }
-    return Future.wait(done);
+    await Future.wait(done);
```
mravn-google: Is this fixing a Dart2 type error or did you intend to change async ordering? The upcoming sync-async change should make the two behave identically as far as the latter goes. (If there is a non-obvious bugfix here, maybe add a test?).

jakobr-google: It's fixing a type error - Future.wait() returns a Future<List<T>>, and we claim to return a Future<Null>. Analyzer doesn't care about this, but the VM in checked mode does at runtime...

Async ordering shouldn't matter much here - we're shutting down the server.

lib/src/client/options.dart
jakobr-google: Fixed ordering. Formatting is spot on according to Dart style guide. :P

PR #61, opened 2018-03-06
Move TLS credentials for server into a separate class.

jakobr-google:
Add address parameter for server, to control which address to listen to,
and the ability to use an ephemeral port.

### DIFF
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -69,10 +69,8 @@ class Server {
     }
   }
 
-  /// Return the port that the server is listening on.
-  ///
-  /// This is only valid when the server is serving, ie. after [serve] has been
-  /// called, and before [shutdown] is called.
+  /// The port that the server is listening on, or `null` if the server is not
+  /// active.
   int get port {
     if (_secureServer != null) return _secureServer.port;
     if (_insecureServer != null) return _insecureServer.port;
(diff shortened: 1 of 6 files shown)


lib/src/server/server.dart
```diff
@@ -23,62 +23,77 @@ import '../shared/security.dart';
...
-    final context = createSecurityContext(true);
-    if (privateKey != null) {
-      context.usePrivateKeyBytes(privateKey, password: privateKeyPassword);
-    }
-    if (certificate != null) {
-      context.useCertificateChainBytes(certificate,
-          password: certificatePassword);
-    }
-    return new Server._(services, port ?? 443, context);
+  /// This is only valid when the server is serving, ie. after [serve] has been
```
mravn-google: Maybe better to document that `null` will be returned, if this getter is called outside the serve-to-shutdown interval. Otherwise, the getter should throw.

lib/src/server/server.dart
```diff
@@ -23,62 +23,77 @@ import '../shared/security.dart';
...
   }
 
-  /// Create a server for the given [services] with no transport security,
-  /// listening on [port].
-  factory Server.insecure(List<Service> services, {int port}) {
-    return new Server._(services, port ?? 80, null);
-  }
-
-  /// Create a secure server for the given [services], listening on [port].
+  /// Return the port that the server is listening on.
```
mravn-google: Per https://www.dartlang.org/guides/language/effective-dart/documentation#prefer-starting-variable-getter-or-setter-comments-with-noun-phrases this should be

/// The port that the server is listening on.

jtattermusch:
[Testing CNCF CLA bot] Please ignore this comment.

PR #75, opened 2018-03-26
Ensure shutdown and terminate always return a Future<Null>.

jakobr-google:
Even if _transport.finish()/terminate() returns some other kind of
Future.


jakobr-google:
cc @munificent

PR #77, opened 2018-04-03
Dart 2 fixes

jakobr-google:
Fixes runtime type errors in Dart 2 mode.

### DIFF
--- a/lib/src/auth/auth.dart
+++ b/lib/src/auth/auth.dart
@@ -72,11 +72,13 @@ abstract class HttpBasedAuthenticator extends BaseAuthenticator {
     return _call;
   }
 
-  Future obtainCredentialsWithClient(http.Client client, String uri);
+  Future<auth.AccessCredentials> obtainCredentialsWithClient(
+      http.Client client, String uri);
 }
 
 class ComputeEngineAuthenticator extends HttpBasedAuthenticator {
-  Future obtainCredentialsWithClient(http.Client client, String uri) =>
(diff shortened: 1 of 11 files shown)


lib/src/auth/auth.dart
```diff
@@ -29,7 +29,7 @@ abstract class BaseAuthenticator {
   auth.AccessToken _accessToken;
   String _lastUri;
 
-  Future authenticate(Map<String, String> metadata, String uri) async {
+  Future<Null> authenticate(Map<String, String> metadata, String uri) async {
```
mravn-google: Can't you use `Future<void>` here yet?

lib/src/server/service.dart
```diff
@@ -34,6 +37,56 @@ class ServiceMethod<Q, R> {
       this.streamingResponse,
       this.requestDeserializer,
       this.responseSerializer);
+
+  StreamController<Q> createRequestStream(StreamSubscription incoming) =>
+      new StreamController(
```
mravn-google: Is `new StreamController<Q>` inferred here?

lib/src/auth/auth.dart
jakobr-google: I'd like to keep this code running on latest stable Dart SDK as long as possible, so not yet...

lib/src/server/service.dart
jakobr-google: I think so - it works, or at least I don't get the runtime exception anymore. :)

I'll make it explicit.

Issue #84, opened 2018-05-01
Dart2 runtime failures in tests


keertip:
running tests in dart 2 mode, see failures like

type 'ArgMatcher' is not a subtype of type 'List<Header>' where...
type 'ArgMatcher' is not a subtype of type 'List<Header>' where
  ArgMatcher is from package:mockito/src/mock.dart
  List is from dart:core
  Header is from package:http2/src/hpack/hpack.dart
../../dart/grpc/test/src/client_utils.dart 126:32  ClientHarness.setUp

keertip:
@mravn-google , @kevmoo

mravn-google:
@jakobr-google

PR #85, opened 2018-05-02
Fix tests in Dart 2 mode.

jakobr-google:
Also updated generated code using latest protoc_plugin, for Dart 2 updates.


jakobr-google:
Fixes #84.

Issue #79, opened 2018-04-04
interceptor support


theobouwman:
How can I add a interceptor to my service?

gedw99:
looks like it just landed 
https://github.com/grpc/grpc-dart/pull/86

mit-mit:
This did not land yet, but we expect it to merge it shortly once we get https://github.com/grpc/grpc-dart/issues/91 resolved

mit-mit:
Version 0.5.0 with this addition has now been published, thanks much to @mogol for the contribution!

PR #86, opened 2018-05-03
Added interceptors.

mogol:
Fixes https://github.com/grpc/grpc-dart/issues/79

---

Draft implementation for interceptors #79  to support analysis of metadata (e.g. authorization)

```
abstract class Interceptor {
  GrpcError handle(ServiceCall call);
}
```

Any feedback is appreciated.


lib/src/server/interceptor.dart
```diff
@@ -0,0 +1,7 @@
+import '../shared/status.dart';
+
+import 'call.dart';
+
+abstract class Interceptor {
+  GrpcError handle(ServiceCall call);
```
sigurdm: We might want to make this async, such that we can do general async handling in the interceptor

thelinuxfoundation:

* One or more commits are not associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub user profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,
CLA GitHub bot_

lib/src/server/interceptor.dart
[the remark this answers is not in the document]
mogol: What is use case? 
What's better way to implement? It could be
* `GrpcError.custom`, check error code  and sink stream without sending the error. 
* Exception, on any exception silently sink stream

mogol: Sounds interesting, looks like it will require a refactoring, old code is fully synchronous.
I tried to return Future<GrpcError> and mark functions async. UT started to fail.
@sigurdm Don't you mind to do it in separate PR?

Should Interceptor return always `Future<GrpcError>` in this case?

mit-mit:
Yes, looks like the test failures are also happening in master. Filed https://github.com/grpc/grpc-dart/issues/91 to track.

lib/src/server/interceptor.dart
```diff
@@ -0,0 +1,13 @@
...
+import 'call.dart';
+import 'service.dart';
+
+/// A gRPC Interceptor.
+///
+/// An interceptor is called before the corresponding [ServiceMethod] invocation.
+/// If the interceptor returns a [GrpcError], the error will be returned as a response and [ServiceMethod] wouldn't be called.
+/// If the interceptor throws [Exception], [GrpcError.internal] with exception.toString() will be returned.
+/// If the interceptor returns null, the corresponding [ServiceMethod] of [Service] will be called.
+typedef Interceptor = GrpcError Function(
```
mogol: @sigurdm I have checked go implementation for interceptors

https://github.com/grpc/grpc-go/blob/a3e99ebee0a498c3093e9741d09c579f0fc34a7f/server.go#L1020

From my understanding it has only two cases - `no-error` and `error`, 
if there is an error, write a status and stop processing a request.
If there is no error, continue processing.

Imo it doesn't make sense to add this feature (I mean silent blocking) now.
Also I am not sure it conforms to grpc standard or not. 

What do you think? Is it required?

sigurdm: No let's stick with error/no-error!

mogol:
Is anything else required to be done in this PR?

sigurdm:
No I don't think so. Thanks!
We need to fix the build issues. Then I'll merge this.

mit-mit:
Merging now that the build is passing. Thanks @mogol for the contribution!

Issue #73, opened 2018-03-24
Remove `version` from every pubspec that is not published


kevmoo:
Help minimize the chance of accidental publishing...

jonkirkman:
#90

PR #90, opened 2018-06-30
#73 removed version from example pubspecs

jonkirkman:
Fixes https://github.com/grpc/grpc-dart/issues/73


mit-mit:
Thanks for contributing @jonkirkman !

PR #98, opened 2018-07-12
Declare Dart 2.0 compatability

mit-mit:


mit-mit:
cc @sortie can you review, please? (just leave a LGTM here in comments)

mit-mit:
cc @kevmoo to confirm that this is the type of change we want to make in general. We tested that this works fine with a 2.0 stable build (except that it still fails as the deps of this package are not updated)

mit-mit:
Moving discussion to https://github.com/dart-lang/protobuf/pulls as this was merged in order for us to test

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.0 - 2018-07-12
```
kevmoo: Avoid breaking change releases for this type of update. Will require folks to update their pubspecs, which is not needed!

PR #103, opened 2018-07-19
Allow for non ascii headers

sigurdm:
Hopefully this will fix https://github.com/grpc/grpc/issues/16004

### DIFF
--- a/lib/src/client/connection.dart
+++ b/lib/src/client/connection.dart
@@ -14,6 +14,7 @@
 // limitations under the License.
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:http2/transport.dart';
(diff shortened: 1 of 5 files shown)


lib/src/client/connection.dart
```diff
@@ -91,7 +92,7 @@ class ClientConnection {
       _userAgent,
     ]);
     metadata?.forEach((key, value) {
-      headers.add(new Header.ascii(key, value));
+      headers.add(new Header(utf8.encode(key), utf8.encode(value)));
```
mkustermann: Maybe we should stick with `ascii.encode(key)` for the key?

sigurdm: Yes - I missed reverting this one!

Issue #43, opened 2017-12-20
Support gRPC-Web protocol


jakobr-google:
See the [gRPC-Web protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) document for details.

We should be able to support this by making a `GrpcWebChannel` that implements the `ClientChannel` interface.

gedw99:
See the improbable git repo for a grpc-web implementation that support normal grpc and grpc with browsers using websockets to give bi-directional streaming.
This version also adheres to the spec.
It also has great interceptor integration for logging, tracing and metric using Prometheus and is heavily production battle tested as they say.


It's golang, not dart based but much easier and flexible than the nginx and cpp way.

jakobr-google:
No-one is talking about nginx or c++ here. This issue is for implementing the gRPC-Web protocol in Dart.

gedw99:
did not know :)

sounds good...

sgon00:
Hi, any updates and plans on this requirement? 

I plan to use grpc/http2 only *(displacement of old http1 REST and websockets)* in a new flutter project. For web interface, I plan to try angulardart v5 because it uses the same dart language. But after some research, I found out there were no native dart grpc-web support. 

So far, I found out:

* The available gRPC servers that exposes services over [gRPC-Web protocol](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) to modern browsers:
    * nginx with gRPC-Web support built from https://github.com/grpc/grpc-web *(private repo)*
    * [a standalone Go binary proxy](https://github.com/improbable-eng/grpc-web/tree/master/go/grpcwebproxy)
    * [caddy server](https://github.com/mholt/caddy) through [grpc plugin](https://github.com/pieterlouw/caddy-grpc) which uses [the above Go package](github.com/improbable-eng/grpc-web/go/grpcweb)
    * [envoy](https://github.com/envoyproxy/envoy) through gRPC-Web http filter
* The available gRPC-Web clients
    * [Javascript and Typescript](https://github.com/improbable-eng/grpc-web/tree/master/ts)
    * [Vue.js](https://github.com/b3ntly/vue-gRPC)
    * [GopherJS](https://github.com/johanbrandhorst/gopherjs-improbable-grpc-web-example)
    * **AngularDart is missing here!!!**

[...]

PR #109, opened 2018-08-05
Support grpc-web in grpc-dart

fuzzybinary:
This is a work in progress, and I haven't gone down the path of implementing grpc-web yet, but I wanted to get some feedback on this refactor while I work on a GrpcWebTransport.

Depending on feedback, I can abandon this PR and create another once that transport is done.

One thing that needs to be done is there are no unit tests for the Http2Transport, only generalized transport.


thelinuxfoundation:

* User @fuzzybinary isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

listepo-alterpost:
Hi, any news?

fuzzybinary:
@listepo-alterpost I've signed the CLA (not sure how often that updates). I've started work on gRPC-Web in a personal branch but no updates yet. Was hoping to get some feedback from someone on whether this is a good approach. 

@kevmoo do you know someone who might be good to reach out to?

fuzzybinary:
To update this (and update #43) I now have very basic gRPC-Web support working, but I have not tried anything but the simplest example in this branch
image

I'd like to get feature parity with the echo example before this can be merged in, but I'm leaving the PR open for feadback.

listepo-alterpost:
@fuzzybinary good job 👍

gedw99:
Using this necessitates TLS 1.3 ?

fuzzybinary:
@gedw99 The HTTP2 stuff might but I don't really know. I haven't tested the gRPC-Web transport against https yet, so it could just fail outright. I'll handle that once I write / fix the unit tests.

kevmoo:
FYI @sigurdm @mit-mit

lib/src/client/transport/web_transport.dart
```diff
@@ -0,0 +1,137 @@
...
+        if(_request.readyState == readyStateDone) {
+          _incomingProcessor.close();
+        }
+      }
+    });
+  }
+
+  @override
+  void terminate() {
+    // TODO: implement terminate
```
nichite: Other than closing the incoming / outgoing streams, is there anything else to do here? My impression was that since we're just delegating to XHR we don't need to do the same cleanup. Maybe call abort() if the request is in flight?

lib/src/client/transport/web_transport.dart
```diff
@@ -0,0 +1,137 @@
...
+  HttpRequest _request;
+  StreamController<ByteBuffer> _incomingProcessor;
+  StreamController<GrpcMessage> _incomingMessages;
+  StreamController<List<int>> _outgoingMessages;
+
+  // TODO: implement id
+  @override
+  int get id => null;
+
+  // TODO: implement incomingMessages
```
nichite: Are these TODOs still needed?

lib/src/client/transport/web_transport.dart
```diff
@@ -0,0 +1,137 @@
...
+      .transform(grpcDecompressor())
+      .listen(_incomingMessages.add,
+        onError: _incomingMessages.addError,
+        onDone: _incomingMessages.close);
+
+    _outgoingMessages.stream
+      .map(GrpcHttpEncoder.frame)
+      .listen((data) {
+        //final encoded = base64Encode(data);
+        _request.send(data);
```
nichite: One thing that's not clear from wrapping the xhr HttpRequest like this: what's the behavior of client / server / bi-directional streaming cases through this transport type? It looks like in client / directional cases, just send the first request?

fuzzybinary: This gets at some of the other comments about the missing implementations of the other calls.

I might need to find a way to determine whether we're making a single call or a request for bi-directional streaming which would require WebSocket support, not an XHR call. At the moment, the XHR should support the single RPC call and a server stream (although it's buggy at the moment) but not client stream or bi-directional stream.

Apparently client / bi-directional streams are in some gRPC-Web implementaitons, so I just have to figure out how to port them.

fuzzybinary:
Thanks for the review - I'll fix some of the nits and the dartfmt issues sometime this week and continue to look into unit tests for the transports, fixing up some of the bugs and looking into websocket support.

fuzzybinary:
Looks like I might have to switch away from HttpRequest to the package:html Request / StreamedResponse APIs in order to support server side streaming.  This also may be easier to transition to WebSockets later on down the road.

gedw99:
@fuzzybinary.

This might help  as it's a 100% golang grpc / grpc server than also produces typescript clients.

https://github.com/improbable-eng/grpc-web

I use this currently and it's been stable for over a year.

It might help answer some of the questions about approaches to bi-directional streaming.

gedw99:
I avoided this one due to many dependencies of nginx etc.

https://github.com/grpc/grpc-web

fuzzybinary:
@gedw99 Awesome thanks! I'd actually looked at that repo before but for some reason it didn't register that their TS is much cleaner. Reading it unblocked me on getting the XHR support working for server side streaming, so now the echo example works.

Thanks!

gedw99:
@fuzzybinary great ! Yeah i can see what you mean that seeing it as Typescript is much easier to understand whats going on.

I have been using that lib with GRPC dart and had no issues btw.
It woudl be great if GRPC-web Dart and GRPC-web golang are interchangeable. 
For example i tend to want to run golang on the server and dart on the clients. 
Some others might want to run Dart on the server and client. 

t would be super productive is this is possible.

fuzzybinary:
That should be 100% possible provided everything conforms to standard. The grpc-dart server should work with any grpc-web client provided there's a grpc-web proxy in front. Similarly, the grpc-dart web client should work with any grpc server provided there's a grpc-web proxy in front.

That said, only the improbable-eng go proxy has experimental support for WebSockets at the moment, and I don't know if that's going to enter the standard or not.

gedw99:
@fuzzybinary websockets can be turned off with this feature flag.

https://github.com/improbable-eng/grpc-web/commit/2cf301900d9d461141af199922f4fe1660ac6022

thelinuxfoundation:

* User @jakemac53 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

fuzzybinary:
I've added some unit tests for both the http2 transport and the XHR transport, so removing the WIP label on this. I think this is ready to go for preliminary grpc-web support, and I'll open another PR for websocket support when I get around to that.

gedw99:
Is there an example in this repo yet ?

fuzzybinary:
In the feature/grpc-web branch on my fork there's an example in example/grpc-web

gedw99:
Thanks Jeff !

fuzzybinary:
Hey @mikeleonard -

If I'm understanding you correctly, you want Cloudflare to host a gRPC server using Cloudflare workers?

gRPC-web is strictly a client web protocol for communication with proxy that can translate the calls to a HTTP2 gRPC server. It's meant to be used by web browser clients since there's no way for a JavaScript client to establish an HTTP2 connection (yet). I don't think there are any proposals for non-HTTP2 gRPC servers.

Since flutter is able to establish an HTTP2 connection, you want it to talk directly to a gRPC HTTP2 server. This library can already do that without this pull request. Whether or not cloudflare is able to host the server, I'm not really sure.

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,152 @@
...
+  StreamController<List<int>> _outgoingMessages;
+
+  @override
+  Stream<GrpcMessage> get incomingMessages => _incomingMessages.stream;
+
+  @override
+  StreamSink<List<int>> get outgoingMessages => _outgoingMessages.sink;
+
+  XhrTransportStream(this._request) {
+    _incomingProcessor = new StreamController();
```
sigurdm: Not sure why this has to be a stream-controller. Isn't it better as a transform?

fuzzybinary:
Thanks Sigurd! I'll get most of this addressed as soon as I can.

As for the two different top level libraries, this would likely require some changes to the dart generator for gRPC, since it assumes it only has to import the top level library. Is there a reason to avoid the conditional import?

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,152 @@
...
+  HttpRequest _request;
+
+  XhrTransport(this.host, this.port, this.options);
+
+  @override
+  Future<void> connect() async {}
+
+  @override
+  Future<void> finish() async {
+    // TODO: implement finish
```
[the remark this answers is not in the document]
fuzzybinary: That's a good question. TBH I'm not 100% sure. It may be okay to remain empty, but I'll have to do more research to figure that out.

example/grpc-web/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Prerequistes
+You will need a clone of the grpc-web repository to run the example server.
+
+You will also need the dart 'webdev' tool, which you can get by running:
+
+```sh
+$ pub global activate webdev
+```
+
```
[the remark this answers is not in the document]
fuzzybinary: Good call -- you need to run the server yes, but I'm thinking I should just delegate those instructions to whatever is in the grpc-web base repository, in case they change.  I will make that portion of the instrucitons more clear though.

example/grpc-web/web/index.html
```diff
@@ -0,0 +1,41 @@
...
+  <script defer src="main.dart.js"></script>
+</head>
+<body>
+  <div class="container">
+    <div class="row" id="first">
+      <div class="form-group">
+        <div class="input-group">
+          <input type="text" class="form-control" id="msg">
+          <span class="input-group-btn">
+            <button class="btn btn-primary" type="button" id="send">Send
```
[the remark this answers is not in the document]
fuzzybinary: Yes. In actuality I should do what the original example does and output responses to the browser window. Will fix.

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,152 @@
...
+
+  @visibleForTesting
+  void initializeRequest(HttpRequest request, Map<String, String> metadata) {
+    for (final header in metadata.keys) {
+      request.setRequestHeader(header, metadata[header]);
+    }
+    request.setRequestHeader('Content-Type', 'application/grpc-web+proto');
+    request.setRequestHeader('X-User-Agent', 'grpc-web-dart/0.1');
+    request.setRequestHeader('X-Grpc-Web', '1');
+    request.overrideMimeType('text/plain; charset=x-user-defined');
```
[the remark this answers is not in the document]
fuzzybinary: Yes -- this and the use of `_request.response as String` from above allow me to process the request streaming instead of processing it after it closes. This is to allow Server Side Streaming requests.

lib/src/client/transport/xhr_transport.dart
fuzzybinary: I'm going to have to go with your judgement there. If you think so, I agree and I'll switch it over.

sigurdm:
True that we would need to update the generator.
Conditional imports are only barely supported currently: The analyzer does not check for consistency between the libraries. It is also harder to test and less modular.

lib/src/client/connection.dart
```diff
@@ -14,16 +14,18 @@
...
 
-import 'package:http2/transport.dart';
 import 'package:meta/meta.dart';
 
-import '../shared/timeout.dart';
-
+import '../shared/status.dart';
 import 'call.dart';
 import 'options.dart';
+import 'transport/http2_transport.dart';
```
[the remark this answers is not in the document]
fuzzybinary: I'm actually not sure why that works either. I've added a stub that tests for dart.library.io just in case.

lib/src/client/transport/transport.dart
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+
+import '../../shared/streams.dart';
```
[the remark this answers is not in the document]
fuzzybinary: I tried this with an 'export' to make the transports available, but that makes the testing imports a bit hairy. I'd rather leave the conditionnal import in the connection if possible.

lib/src/client/transport/xhr_transport.dart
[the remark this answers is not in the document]
fuzzybinary: 👍 Done!

fuzzybinary:
I think I hit most of your concerns. Let me know if there's anything else!

fuzzybinary:
Concerning the conditional imports, I'm wondering if it makes sense to push this through for now, then change the generator to accept a transport moving forward and adjust this code accordingly when that goes through.

lib/src/client/transport/xhr_transport_stub.dart
```diff
@@ -0,0 +1,45 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+
+import '../options.dart';
```
nichite: I believe the transitive dependency on dart:io will make this stub moot

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,144 @@
...
+    await _incomingProcessor.close();
+    await _outgoingMessages.close();
+    _request.abort();
+  }
+}
+
+class XhrTransport extends Transport {
+  final String host;
+  final int port;
+  final ChannelOptions options;
```
nichite: Is this ever used? May be able to remove it (see other comment)

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,144 @@
...
+      if (_request.readyState == HttpRequest.DONE) {
+        _incomingProcessor.close();
+        _outgoingMessages.close();
+      }
+    });
+
+    _request.onProgress.listen((data) {
+      // use response over responseText as most browsers don't support
+      // using responseText during an onProgress event.
+      final responseString = _request.response as String;
```
nichite: Is `_request.response` just the incremental chunk of the response, or the entire response so far?

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,144 @@
...
+        }
+      }
+
+      if (_request.readyState == HttpRequest.DONE) {
+        _incomingProcessor.close();
+        _outgoingMessages.close();
+      }
+    });
+
+    _request.onProgress.listen((data) {
```
nichite: can swap `data` with `_` if unused

lib/src/client/options.dart
```diff
@@ -51,6 +51,10 @@ typedef bool BadCertificateHandler(X509Certificate certificate, String host);
 /// certificates, etc.
 bool allowBadCertificates(X509Certificate certificate, String host) => true;
 
+/// Transport type. Use Http2 for vm whenever possible, GrpcWeb in browsers that do
+/// not support Http2
+enum TransportType { Http2, Xhr, Websocket }
```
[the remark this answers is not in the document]
fuzzybinary: Agreed that could be better, though the question later becomes being able to choose betwen Xhr and Websocket transport on Web.

One thing that we should keep in mind is that considering the two top-level lib issue is that we do want to support a shared code use case -- using a shared logic library for Angular and Flutter. So long as we know where that clear division is I think we'll be fine.

lib/src/client/options.dart
```diff
@@ -93,15 +97,18 @@ class ChannelCredentials {
...
 
   const ChannelOptions(
       {ChannelCredentials credentials,
       Duration idleTimeout,
+      TransportType transportType = TransportType.Http2,
       BackoffStrategy backoffStrategy =
           defaultBackoffStrategy}) // Remove when dart-lang/sdk#31066 is fixed.
       : this.credentials = credentials ?? const ChannelCredentials.secure(),
         this.idleTimeout = idleTimeout ?? defaultIdleTimeout,
+        this.transportType = transportType ?? transportType,
```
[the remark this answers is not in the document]
fuzzybinary: Yeah, that's weird. Will fix.

lib/src/client/transport/xhr_transport_stub.dart
fuzzybinary: So, interestingly, dart:io imports okay on web... you just can't do anything with it. I'm curious why that is... ;)

lib/src/client/transport/xhr_transport.dart
fuzzybinary: It's the entire response recieved so far. It's find for short streaming sessions but definately one of the reasons we would not want to use Xhr for server side streaming calls long term.

lib/src/client/options.dart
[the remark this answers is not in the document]
fuzzybinary: So interesting dart related question -- technically even with default parameters, a caller could pass in "null" for any of these parameters.  Is it better to be safe and check for null or just use default parameters? Is there a "notnull" meta annotation (required is the closest I can find but not quite right).

lib/src/client/transport/http2_transport_stub.dart
```diff
@@ -0,0 +1,45 @@
...
+import '../options.dart';
+
+import 'transport.dart';
+
+class Http2Transport extends Transport {
+  Http2Transport(String host, int port, ChannelOptions options);
+
+  @override
+  Future<void> connect() {
+    return Future.value();
```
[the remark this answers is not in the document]
fuzzybinary: Good call, will fix.

lib/src/client/options.dart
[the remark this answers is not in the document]
nichite: That's right, I think the majority of the code can be in a third "core" library. The way I was figuring, the gRPC HTTP2-specific lib can depend on core and add Http2Transport, etc. gRPC-web lib can depend on core and add XhrTransport, etc. As much logic as possible should be shared, which should be the case if we can get `call.dart` and `connection.dart` able to import on both platforms.

lib/src/client/options.dart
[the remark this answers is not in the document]
nichite: That's right, if you explicitly set a named parameter to null, it'll override the default value. There isn't a @notNull annotation available for public use--there is one that I believe can only be accessed internally by core Dart libraries.

Ultimately, I think it's up to the specific API owner to determine whether or not to allow null values, depending on what's going on. Generally though setting null to signify some special case or absent is a code smell that can be resolved by using [Optional<T>](https://github.com/google/quiver-dart/blob/master/lib/src/core/optional.dart) instead. But yeah, when I know something needs to be set I'll use [checkNotNull](https://github.com/google/quiver-dart/blob/54f6693e842c607ec4c9346c5003b4ce8f9e6c4d/lib/check.dart#L61) in the constructor.

It could very be (though unlikely) that the original author used parameter defaults to specify nullable fields and the initializers via '??' for non-nullable ones, though if that were the case it should have been documented. For TransportType in particular, I feel like we should enforce that it's set, since just blindly guessing a default value for Flutter or web will be wrong a lot.

Vaguely related: I do know that the built_value library goes the other way and uses  a "nullable" [annotation](https://github.com/google/built_value.dart/blob/c5ea1df7ae73bf3550604e5ab3c3b23b00cdce04/built_value/lib/built_value.dart#L90), though that's for a very specific type of data class.

lib/src/client/transport/xhr_transport_stub.dart
nichite: Yup that's weird. It could be the strict dependency platform checks don't work the same externally as I'm used to, but we definitely shouldn't rely on that behavior. Shouldn't matter if we're splitting this up, in either case.

lib/src/client/transport/xhr_transport.dart
fuzzybinary: I looked into this (forgot it in the initial update) but because I need to add to the stram from onProgress, I think this still makes sense as a stream controller, even though it's piping out to a secondary stream.

It is possible I'm just not sure how to properly implement it as a transformer though.

fuzzybinary:
Updated. Let me know if you have any more comments.

sigurdm:
I think we have to decide on the library structure already from this pr. We should minimize breakage unless we have a good reason

fuzzybinary:
@sigurdm I was not able to reproduce by just using "30 test" on my box. Is this with a DDC build or a dart2js build? Are there any other steps?

sigurdm:
No, I just I run the example. This is DDC. (webdev serve).

fuzzybinary:
Weird. Linux? Mac? Windows? Which version of Chrome? I can't repro unforutnately...

sigurdm:
Linux, chrome Version 70.0.3538.67 (Official Build) (64-bit).
I get the error: 
```
POST http://localhost:8080/grpc.gateway.testing.EchoService/ServerStreamingEcho net::ERR_INCOMPLETE_CHUNKED_ENCODING 200 (OK)
Uncaught Error: RangeError: Value not in range: 1200
    at Object.wrapException (js_helper.dart:1763)
    at Interceptor.substring$2 (js_string.dart:160)
    at Interceptor.substring$1 (js_string.dart:155)
    at Object.J.substring$1$s (main.dart.js:12556)
    at XhrTransportStream_closure1.call$1 (xhr_transport.dart:84)
    at _EventStreamSubscription_closure.call$1 (html_dart2js.dart:35678)
    at invokeClosure (js_helper.dart:2437)
    at XMLHttpRequest.<anonymous> (js_helper.dart:2471)
```
I get a slightly different message on firefox.
I could reproduce with dart2js also.
It seems to be a 15s timeout, if I reduce the time between the messages I get more messages through before the failure.

I think tinkering with the HTTPRequest.timeout might have some effect.

fuzzybinary:
I just updated to Chrome 70 and can now reproduce on Mac. I'll look into it and see what I can see.

lib/src/client/connection.dart
```diff
@@ -14,17 +14,19 @@
...
-import 'package:http2/transport.dart';
 import 'package:meta/meta.dart';
 
-import '../shared/timeout.dart';
-
+import '../shared/status.dart';
 import 'call.dart';
 import 'options.dart';
 
+import 'transport/http2_transport_stub.dart'
```
eredo: For me this wasn't buildable using `build_runner`. The following error occurred: 

```
[SEVERE] build_web_compilers|ddc on package:grpc/grpc.dartdevc.module:
Error compiling dartdevc module:grpc|lib/grpc.ddc.js

[error] Target of URI doesn't exist: 'transport/http2_transport.dart'. (package:grpc/src/client/connection.dart, line 25, col 26)
[error] Target of URI doesn't exist: 'transport/xhr_transport_stub.dart'. (package:grpc/src/client/connection.dart, line 27, col 8)

Please fix all errors before compiling (warnings are okay).
}
```

Weirdly I was able to fix this by adding "ignore: uri_does_not_exist". I saw this approach within the `http` package.

```
// ignore: uri_does_not_exist
import 'transport/http2_transport_stub.dart'
    // ignore: uri_does_not_exist
    if (dart.library.io) 'transport/http2_transport.dart';
import 'transport/transport.dart';
// ignore: uri_does_not_exist
import 'transport/xhr_transport_stub.dart'
// ignore: uri_does_not_exist
    if (dart.library.html) 'transport/xhr_transport.dart';
```

Any idea why this is necessary?

Brahmasmi:
Based on my limited understanding, there is a question of whether the refactor should happen in this PR or a subsequent one. I also understand that there are external changes in the dart generator which would be required for that to happen. Additionally, it is apparent from the discussion that there are better checks and tooling for Dart available within Google.

In my layman opinion, I think it would be preferable if this PR gets refactored and merged along with the generator changes.

@fuzzybinary @nichite @sigurdm @eredo What do you think?

sigurdm:
Yes - If I did not make it clear before, I think the refactor should happen before landing. Branches are for trying things out, master is where we land what we want to have.

Brahmasmi:
Tak Sigurd.

So, for this PR to land, it will have to be refactored.

Now, would I be correct if I say that @nichite volunteered to work on the generator to speed things up. Does that offer still stand Nic?

Also, @fuzzybinary would you be in a position to refactor, assuming Nic is able to patch the generator? Would you need collaboration / help from within Google? Please feel free to ask/decline.

The idea is to see this land and released.

Dhanyavaad.

fuzzybinary:
@Brahmasmi I'm still happy to help. I'm currently a bit swamped so might not be able to pick back up until Dec.

Are the patches to the genertator expected to modify how this library needs to work? Do I need to wait for those to land before continuing on implementing this feature?

I just need to understand constraints / whether we should aim for no breaking changes (there's a bit of back and forth on that in the PR right now).

Brahmasmi:
@fuzzybinary Thank you for your offer of continuing help.

Would it be possible for the Dart team to collaborate with @fuzzybinary, co-ordinate the generator changes and effectively get this PR landed? In other words, own this?

I ask this because this functionality is, if I am not mistaken, one of the roadblocks towards increased adoption of AngularDart. Please feel free to peruse the number of thumbs up as well as recent comments on issue #43 .

Dhanyavaad.

fuzzybinary:
Looks like the timeout bug is on the envoy side - it times out a stream after 15 seconds without further configuration.
https://github.com/grpc/grpc-web/issues/309

I plan on adding a check to prevent the dart exception in the case that the stream is closed pre-maturely.

sigurdm:
This has not been completely forgotten - just many other things are going on.
I plan to put some effort into getting this landed early q1.

fuzzybinary:
Let me know if there's anything I can do to help!

sigurdm:
I started working on updating the generated code to not depend on dart:io.
That should enable us to do a cleaner split.

fuzzybinary:
@sigurdm Sorry for the delay - I've approved and merge your changes into my fork so this should be ready to go.

supermuka:
I'm happy and I think many other Dart web enthusiasts are also seeing that soon we will have a published version of grpc-dart for web. As a suggestion if possible, in addition to the `example\grpc-web` example containing a `node` implementation we also have an example with Envoy + grpc implementation in Dart Server Side (vm). I'm not sure how to configure this yet!

.travis.yml
```diff
@@ -1,14 +1,19 @@
...
+  
+# Run against both the dev and stable channel.
 dart:
   - stable
   - dev
 
 # Define test tasks to run.
 dart_task:
   - test: --platform vm
+  - test: --platform chrome
```
jonasfj: I'm not sure if it's easy to do, but it might be worth testing with Firefox too..

I would love to add safari and edge, but I understand if those are hard to automate :)

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,140 @@
...
+        .listen((data) => _request.send(data));
+
+    _request.onReadyStateChange.listen((data) {
+      final contentType = _request.getResponseHeader('Content-Type');
+      if (contentType == null) {
+        return;
+      }
+
+      if (_request.readyState == HttpRequest.HEADERS_RECEIVED) {
+        if (contentType.startsWith('application/grpc')) {
```
jonasfj: Are we completely unconcerned with the response status?

lib/src/client/transport/xhr_transport.dart
```diff
@@ -0,0 +1,140 @@
...
+          }
+
+          // Force a metadata message with headers
+          final headers = GrpcMetadata(_request.responseHeaders);
+          _incomingMessages.add(headers);
+        }
+      }
+
+      if (_request.readyState == HttpRequest.DONE) {
+        _incomingProcessor.close();
```
jonasfj: What if `_request.onError` happened? Perhaps the connection was broken... If this was a streaming response would we know the difference between a broken connection and a connection that closed because no more messages was supposed to arrive?

lib/src/client/connection.dart
```diff
@@ -46,98 +45,38 @@ enum ConnectionState {
...
 
-  ClientTransportConnection _transport;
+  Transport _transport;
 
   /// Used for idle and reconnect timeout, depending on [_state].
   Timer _timer;
   Duration _currentReconnectDelay;
 
-  ClientConnection(this.host, this.port, this.options);
+  ClientConnection(this.host, this.port, this.options, this.connectTransport);
```
jonasfj: Won't this break backwards compatibility?

It probably wasn't a good idea for `ClientConnection` to be public concrete type in the first place :)
Since we probably don't want to people to instantiate their own `ClientConnection` objects and return them from a custom `ClientChannel` implementation.

jonasfj:
@sigurdm, feel free to moderate/overrule my suggestions, and make an executive decision to land this..
I had one concern regarding `xhr2` error handling, but otherwise it's mostly about the public API, and whether we can call this backwards compatible :)

Cleaning up the public API could be done separately. Even if you decide that this needs a major version  bump.

sigurdm:
Thanks a bunch for the review @jonasfj , and for all the patience @fuzzybinary. I will merge this now (but not publish yet), and do a follow up PR addressing the concerns.

vjpai:
This pull request is breaking interop tests at grpc/grpc . Errors seem to be of the form:

```
2019-03-16 16:01:13,286 file:///var/local/git/grpc-dart/lib/src/client/http2_channel.dart:24:1: Error: 'Future' expects 0 type arguments.
Future<Transport> connectTransport(
^
file:///var/local/git/grpc-dart/lib/src/client/http2_channel.dart:26:10: Error: A value of type '#lib1::Http2Transport' can't be assigned to a variable of type 'dart.async::FutureOr<invalid-type>'.
Try changing the type of the left hand side, or casting the right hand side to 'dart.async::FutureOr<invalid-type>'.
  return Http2Transport(host, port, options)..connect();
         ^
file:///var/local/git/grpc-dart/lib/src/client/http2_channel.dart:24:19: Error: Functions marked 'async' must have a return type assignable to 'Future'.
Future<Transport> connectTransport(
                  ^^^^^^^^^^^^^^^^
file:///var/local/git/grpc-dart/lib/src/client/http2_channel.dart:32:21: Error: The argument type '(dart.core::String, dart.core::int, #lib1::ChannelOptions) → invalid-type' can't be assigned to the parameter type '(dart.core::String, dart.core::int, #lib1::ChannelOptions) → dart.async::Future<#lib2::Transport>'.
Try changing the type of the parameter, or casting the argument to '(dart.core::String, dart.core::int, #lib1::ChannelOptions) → dart.async::Future<#lib2::Transport>'.
      : super(host, connectTransport, port: port, options: options);
                    ^

```

Any ideas of what we can do here? @jtattermusch @nicolasnoble

sigurdm:
Yes - this is landed incomplete because it was becoming too difficult to manage. But the error you see is because your dart SDK is before v. 2.1

sigurdm:
I will move this to a branch, and finish it there.

nicolasnoble:
@sigurdm we don't really have much knowledge on our side about dart, the dart tests we have in there were originated by the grpc-dart authors. Would it be possible to send a PR on grpc/grpc to do whatever change is needed to fix this?

sigurdm:
The change is now reverted (until we make it work better). Your tests should start working again.
Thanks for following up.

Issue #108, opened 2018-08-05
The method 'remove' was called on null.


Kleak:
grpc with latest version from pub.

i got this error sometimes but not sure right now how to reproduce exactly.

```
Unhandled exception:
NoSuchMethodError: The method 'remove' was called on null.
Receiver: null
Tried calling: remove(":status")
#0      Object.noSuchMethod (dart:core/runtime/libobject_patch.dart:46:5)
#1      ServerHandler.sendTrailers (package:grpc/src/server/handler.dart:286:23)
#2      ServerHandler._sendError (package:grpc/src/server/handler.dart:350:5)
#3      ServerHandler._onResponse (package:grpc/src/server/handler.dart:239:7)
#4      _RootZone.runUnaryGuarded (dart:async/zone.dart:1314:10)
#5      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:336:11)
#6      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:263:7)
#7      _SyncStreamController._sendData (dart:async/stream_controller.dart:763:19)
#8      _StreamController._add (dart:async/stream_controller.dart:639:7)
#9      new Stream.fromFuture.<anonymous closure> (dart:async/stream.dart:107:18)
#10     _RootZone.runUnary (dart:async/zone.dart:1379:54)
#11     _FutureListener.handleValue (dart:async/future_impl.dart:129:18)
#12     Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:642:45)
#13     Future._propagateToListeners (dart:async/future_impl.dart:671:32)
#14     Future._complete (dart:async/future_impl.dart:476:7)
#15     _SyncCompleter.complete (dart:async/future_impl.dart:51:12)
#16     _completeOnAsyncReturn (dart:async/runtime/libasync_patch.dart:292:13)
#17     IngredientService.search (package:backend/src/services/ingredient/ingredient.dart)
<asynchronous suspension>
#18     IngredientServiceBase.search_Pre (package:protos/src/generated/ingredient.pbgrpc.dart:139:12)
<asynchronous suspension>
#19     ServiceMethod.handle (package:grpc/src/server/service.dart:63:20)
#20     ServerHandler._startStreamingRequest (package:grpc/src/server/handler.dart:153:30)
#21     ServerHandler._onDataIdle (package:grpc/src/server/handler.dart:129:5)
#22     _RootZone.runUnaryGuarded (dart:async/zone.dart:1314:10)
[...]

PR #120, opened 2018-09-07
allows async interceptors

a14n:


a14n:
ffe489b could perhaps fix #108 as I faced the same error.

lib/src/server/handler.dart
```diff
@@ -146,7 +148,6 @@ class ServerHandler extends ServiceCall {
   }
 
   void _startStreamingRequest() {
-    _incomingSubscription.pause();
```
sigurdm: Is this change related to the interceptors?

a14n: It's related to async interceptor (and perhaps more as #108 reports the same error I have before the last commit ffe489b)

a14n:
Could this PR be merged?

sigurdm:
Yes, sorry for letting you wait.

a14n:
no problem :) Thanks

PR #128, opened 2018-10-17
don't throw on closing channel not open

a14n:
If channel is not open calling `terminate()` or `shutdown` triggered Error:

```
Unhandled exception:
NoSuchMethodError: The method 'shutdown' was called on null.
Receiver: null
Tried calling: shutdown()
#0      Object.noSuchMethod (dart:core/runtime/libobject_patch.dart:46:5)
#1      ClientChannel.shutdown (package:grpc/src/client/channel.dart:50:24)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.6.8
+
+* Calling `terminate()` or `shutdown()` on a channel doesn't throw error if the
+channel is not yet open.
+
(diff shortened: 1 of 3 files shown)


mit-mit:
@sigurdm did you get a chance to look at this?

a14n:
>Would you mind writing a changelog notice and a pubspec update?

Done

PR #132, opened 2018-10-19
handle stream cancel on client side

### LINKED ISSUE #131: Unhandled exception when stream is cancelled
I use `grpc-0.6.5` with the following code:

```dart
main(List<String> args) async {
  await Server([Service()]).serve(port: 8022);

  final client = ServiceClient(ClientChannel(
    'localhost',
    port: 8022,
    options: const ChannelOptions(credentials: ChannelCredentials.insecure()),
  ));

  print(await client.method(Empty()).isEmpty);
  print(await client.method(Empty()).isEmpty); // line 18
}

class Service extends ServiceBase {
  @override
  Stream<Empty> method(ServiceCall call, Empty request) async* {
    yield request;
    yield request;
  }
}
```

and the proto:

```proto
syntax = "proto3";
import "google/protobuf/empty.proto";
service Service {
  rpc Method (google.protobuf.Empty) returns (stream google.protobuf.Empty);
}
```

When I run the code (with `Dart-2.0.0`) I get the following output:

```
false
Unhandled exception:
gRPC Error (14, Error making call: Bad state: The http/2 connection is no longer active and can therefore not be used to make new streams.)
#0      main (file:///home/me/sample/bin/issue-grpc-stream.dart:18:9)
<asynchronous suspension>
#1      _startIsolate.<anonymous closure> (dart:isolate/runtime/libisolate_patch.dart:277:32)
#2      _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:165:12)
Exited (255)
```

Basically the code starts a server that provides a stream method. I call the method 2 times and use a `.isEmpty` on the resulting stream.

It looks like the problem is about cancelling a stream (because `.isEmpty` cancel its subscription directly after receiving the first event).

(I had a hard time to find what was the issue and I think it's an important issue)

a14n:
Fix #131

### DIFF
--- a/lib/src/client/call.dart
+++ b/lib/src/client/call.dart
@@ -155,7 +155,7 @@ class ClientCall<Q, R> implements Response {
       }
       _responses.onPause = _responseSubscription.pause;
       _responses.onResume = _responseSubscription.resume;
-      _responses.onCancel = _responseSubscription.cancel;
+      _responses.onCancel = _safeTerminate;
     }
   }
 


a14n:
:-( actually there's still a bug

sigurdm:
Can you explain more?

a14n:
In https://github.com/grpc/grpc-dart/issues/131#issue-371585506 I put a sample to demonstrate an issue when a stream is canceled. With this patch the sample works fine. But after trying my patch I see now an error (the server crashed) . A simple repro is to use the same sample I mentioned but doing 3 yield in Service. I get the following output:

```
Unhandled exception:
gRPC Error (14, Error making call: Bad state: The http/2 connection is no longer active and can therefore not be used to make new streams.)
dart:async/stream_controller.dart 594:43                      _StreamController.addError
package:grpc/src/client/call.dart 71:18                       ClientCall._terminateWithError
package:grpc/src/client/call.dart 118:7                       ClientCall._sendRequest
package:grpc/src/client/call.dart 92:7                        ClientCall.onConnectionReady
package:grpc/src/client/connection.dart 172:10                ClientConnection._startCall
package:grpc/src/client/connection.dart 150:9                 ClientConnection.dispatchCall
package:grpc/src/client/channel.dart 76:18                    ClientChannel.createCall.<fn>
```

This issue drives me crazy and I'm blocked because of it.

a14n:
Help would be **really** appreciated. :)

PR #146, opened 2018-12-20
Sync internal repo changes.

nichite:
Replace usage of `Future<Null>` with `Future<void>`.

This is a large-ish but trivial change. Using `Future<void>` is preferred to `Future<Null>`.

In addition, the `FutureOr<Null>` typedef for MetadataProviders actually breaks when defining a function that returns `void`. Since `FutureOr<T>` functions can return `Future<T>` or `T`, a provider that doesn't return a future would have to explicitly return `null`. This change should allow metadata provider functions to return `void` and `Future<void>`.


thelinuxfoundation:

* User @nichite isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

nichite:
So this is actually my first extrernal PR--do I need to make another commit to rerun the cla/linuxfoundation check?

PR #152, opened 2019-01-22
Add service_api.dart

sigurdm:
This is a prelude to being able to add grpc-web support without the need for conditional imports.

It will need to be followed up with an update to the protoc_plugin.


sigurdm:
cc @nichite @fuzzybinary

PR #156, opened 2019-03-07
don't send anything if connection have been closed

### LINKED ISSUE #147: Clients that fail to close connections can crash servers
Following https://github.com/grpc/grpc-dart/issues/111, I find that a streaming client that fails to close a connection when it is killed can cause a Dart grpc server to crash. 

It appears that #111 was closed because well-written clients can now be closed safely, but when I tested with a client that did not call `channel.terminate();` on exit, I got the same crash and stack trace as reported in #111. This occurred with both the released (1.0.0) grpc library and on today's `master`.

## Steps to reproduce

pubspec.yaml:
```
name: sample

environment:
  sdk: '>=2.0.0 <3.0.0'

dependencies:
  async: '>=1.13.3 <3.0.0'
  grpc: ^1.0.0
  protobuf: ^0.10.1
  args: ^1.5.1

dev_dependencies:
  test: ^1.3.0
```

protos/sample.proto:
```
syntax = "proto3";

package bug;

message Empty {}

message Hello {
    string hello = 1;
}

service BugService {
    rpc infiniteStream (Empty) returns (stream Hello) {}
}
```

bin/server.dart:
```
import 'dart:async';
import 'package:grpc/grpc.dart' as grpc;
import 'package:sample/src/generated/sample.pb.dart';
import 'package:sample/src/generated/sample.pbgrpc.dart';

class BugService extends BugServiceBase {
  final infiniteStreamController = StreamController<Hello>.broadcast();
  int count = 0;

  BugService() {
    Timer.periodic(Duration(seconds: 1), (_) {
      count++;
      infiniteStreamController.add(Hello()..hello = 'Hello $count');
    });
  }

  @override
  Stream<Hello> infiniteStream(grpc.ServiceCall call, Empty request) async* {
    await for (final hello in infiniteStreamController.stream) {
      yield hello;
    }
  }
}

main(List<String> args) async {
  // Run the gRPC server.
  final server = new grpc.Server([new BugService()]);
  await server.serve(port: 8080);
  print('Server listening on port ${server.port}...');
}
```

bin/client.dart:
```
import 'dart:async';
import 'dart:io';

import 'package:args/args.dart';
import 'package:grpc/grpc.dart';
import 'package:sample/src/generated/sample.pb.dart';
import 'package:sample/src/generated/sample.pbgrpc.dart';

main(List<String> args) async {
  var parser = new ArgParser();
  parser.addFlag('shutdown', defaultsTo: true);
  var results = parser.parse(args);
  var shutdown = results['shutdown'];

  final channel = ClientChannel(
    'localhost',
    port: 8080,
    options: const ChannelOptions(
      credentials: const ChannelCredentials.insecure(),
    ),
  );
  final client = BugServiceClient(channel);
  final completer = Completer<void>();
  final subscription = client.infiniteStream(Empty()).listen(print);
  if (shutdown) {
    ProcessSignal.sigint.watch().listen((_) async {
      await subscription.cancel();
      print('shutting down gracefully');
      await channel.terminate();
      completer.complete();
    });
  }
  await completer.future;
  exit(0);
}
```
Running the server with `dart bin/server.dart` and the client with `dart bin/client.dart`, killing the client with Control-C proceeds successfully because the client listens for SIGINT and closes the channel. But if the client is run with `dart bin/client.dart --no-shutdown` and then killed, the server crashes with the following trace:

```
Unhandled exception:
Bad state: Cannot add event after closing
#0      _StreamController.add (dart:async/stream_controller.dart:584:24)
#1      _StreamSinkWrapper.add (dart:async/stream_controller.dart:858:13)
#2      TransportStream.sendHeaders (package:http2/transport.dart:241:10)
#3      ServerHandler.sendTrailers (package:grpc/src/server/handler.dart:303:13)
#4      ServerHandler._sendError (package:grpc/src/server/handler.dart:352:5)
#5      ServerHandler._onResponse (package:grpc/src/server/handler.dart:241:7)
#6      _RootZone.runUnaryGuarded (dart:async/zone.dart:1314:10)
#7      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:336:11)
#8      _DelayedData.perform (dart:async/stream_impl.dart:591:14)
#9      _StreamImplEvents.handleNext (dart:async/stream_impl.dart:707:11)
#10     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:667:7)
#11     _microtaskLoop (dart:async/schedule_microtask.dart:41:21)
#12     _startMicrotaskLoop (dart:async/schedule_microtask.dart:50:5)
#13     _Timer._runTimers (dart:isolate/runtime/libtimer_impl.dart:391:30)
#14     _Timer._handleMessage (dart:isolate/runtime/libtimer_impl.dart:416:5)
#15     _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:171:12
```

## Expected results

While it seems good to encourage clients to be well-behaved, I don't think we want a badly-written or malicious client to be able to crash a server by breaking a connection incorrectly.

(Is there something incorrect about the way the test server is written?)

a14n:
Fixes #147

### DIFF
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -103,9 +103,13 @@ class Server {
       _connections.add(connection);
       // TODO(jakobr): Set active state handlers, close connection after idle
       // timeout.
-      connection.incomingStreams.listen(serveStream, onError: (error) {
+      final handlers = <ServerHandler>[];
+      connection.incomingStreams.listen((stream) {
+        handlers.add(serveStream(stream));
+      }, onError: (error) {
         print('Connection error: $error');
       }, onDone: () {
+        handlers.forEach((e) => e.isConnectionClosed = true);
         _connections.remove(connection);
       });
     }, onError: (error) {
@@ -113,8 +117,8 @@ class Server {
     });
   }
 
-  void serveStream(ServerTransportStream stream) {
-    new ServerHandler(lookupService, stream, _interceptors).handle();
+  ServerHandler serveStream(ServerTransportStream stream) {
+    return new ServerHandler(lookupService, stream, _interceptors)..handle();
   }
 
   Future<void> shutdown() async {
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -103,13 +103,13 @@ class Server {
       _connections.add(connection);
       // TODO(jakobr): Set active state handlers, close connection after idle
       // timeout.
-      final handlers = <ServerHandler>[];
+      var isConnectionClosed = false;
       connection.incomingStreams.listen((stream) {
-        handlers.add(serveStream(stream));
+        serveStream(stream, () => isConnectionClosed);
       }, onError: (error) {
         print('Connection error: $error');
       }, onDone: () {
-        handlers.forEach((e) => e.isConnectionClosed = true);
+        isConnectionClosed = true;
         _connections.remove(connection);
       });
     }, onError: (error) {
@@ -117,8 +117,9 @@ class Server {
     });
   }
 
-  ServerHandler serveStream(ServerTransportStream stream) {
-    return new ServerHandler(lookupService, stream, _interceptors)..handle();
+  void serveStream(ServerTransportStream stream, bool isConnectionClosed()) {
+    new ServerHandler(lookupService, stream, isConnectionClosed, _interceptors)
+        .handle();
   }
 
   Future<void> shutdown() async {
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -54,9 +54,9 @@ class ServerHandler extends ServiceCall {
   bool _isTimedOut = false;
   Timer _timeoutTimer;
 
-  bool isConnectionClosed = false;
+  final bool Function() isConnectionClosed;
 
-  ServerHandler(this._serviceLookup, this._stream,
+  ServerHandler(this._serviceLookup, this._stream, this.isConnectionClosed,
       [this._interceptors = const <Interceptor>[]]);
 
   DateTime get deadline => _deadline;
@@ -258,7 +258,7 @@ class ServerHandler extends ServiceCall {
   }
 
   void sendHeaders() {
-    if (isConnectionClosed) return;
+    if (isConnectionClosed()) return;
 
     if (_headersSent) throw new GrpcError.internal('Headers already sent');
 
@@ -281,7 +281,7 @@ class ServerHandler extends ServiceCall {
   }
 
   void sendTrailers({int status = 0, String message}) {
-    if (isConnectionClosed) return;
+    if (isConnectionClosed()) return;
 
     _timeoutTimer?.cancel();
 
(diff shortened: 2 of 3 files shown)


a14n:
The issue comes from triing to write when connection has been closed.

I'm not really sure if this should be handle here or in http2 package. /cc @mkustermann

a14n:
ping (strangely I never received notification from this PR)

mkustermann:
@athomas Does anyone on the server team to have cycles for looking at this?

lib/src/server/server.dart
```diff
@@ -103,18 +103,22 @@ class Server {
       _connections.add(connection);
       // TODO(jakobr): Set active state handlers, close connection after idle
       // timeout.
-      connection.incomingStreams.listen(serveStream, onError: (error) {
+      final handlers = <ServerHandler>[];
```
mkustermann: This could be like a memory leak: There can be many thousands of requests being handled via the same http/2 connection. This list might accumulate all of them even if the actual handlers have finished a long time ago.

a14n: Indeed, thanks!

I updated the code to avoid this list. PTAL

a14n:
@sigurdm any chance you take a look at this PR?

sigurdm:
I have started looking into this together with @jonasfj.
We came to the conclusion that the handler ideally should see an error (such that any finally handlers would run) that would then be caught by the framework.

a14n:
I'm not sure to understand your proposal enough to make the changes you expect. Will you send another PR to fix the original issue?

sigurdm:
The issue is that:
```
import 'dart:async';
import 'dart:isolate';
import 'package:grpc/grpc.dart' as grpc;
import 'package:sample/src/generated/sample.pb.dart';
import 'package:sample/src/generated/sample.pbgrpc.dart';


client(_) async {
  final channel = grpc.ClientChannel(
    'localhost',
    port: 8080,
    options: const grpc.ChannelOptions(
      credentials: const grpc.ChannelCredentials.insecure(),
    ),
  );
  final client = BugServiceClient(channel);
  client.infiniteStream(Empty()).listen(print);

}


class BugService extends BugServiceBase {
  final infiniteStreamController = StreamController<Hello>.broadcast();

  BugService() {
  }

  @override
  Stream<Hello> infiniteStream(grpc.ServiceCall call, Empty request) async* {
    int count = 0;
    while(true) {
      count++;
      print("count $count");
      yield Hello()..hello = 'Hello $count';
      await Future.delayed(Duration(milliseconds:100));
    }
  }
}

spawnClient() async {
  Isolate i = await Isolate.spawn(client, null);
  await Future.delayed(Duration(seconds: 1));
  i.kill();
}

main(List<String> args) async {
  // Run the gRPC server.
  final server = new grpc.Server([new BugService()]);

  spawnClient();
  await server.serve(port: 8080);
  print('Server listening on port ${server.port}...');

}

```

Would have the server continue the infinite loop.

I'm trying to come up with something. I think I'm almost there. Will continue Monday.

a14n:
superseded by #164

PR #160, opened 2019-03-21
Handle request errors

sigurdm:


sigurdm:
cc jonasfj

sigurdm:
Good point. I will follow up with a PR trying to minimize what we expose.

PR #164, opened 2019-04-01
Cancel `ServerHandler` when `ServerTransportConnection.incomingStreams` is closed  

sigurdm:
I take the chance to also deprecate ServerHandler and Server.serveStream.
This should fix: https://github.com/grpc/grpc-dart/issues/147


sigurdm:
cc @jonasfj @mkustermann @a14n

lib/src/server/server.dart
```diff
@@ -101,20 +102,35 @@ class Server {
...
       // TODO(jakobr): Set active state handlers, close connection after idle
       // timeout.
-      connection.incomingStreams.listen(serveStream, onError: (error) {
+      connection.incomingStreams.listen((stream) {
+        handler = serveStream_(stream);
+      }, onError: (error) {
         print('Connection error: $error');
       }, onDone: () {
+        // TODO(sigurdm): This is not correct behavior in the presence of
+        // half-open tcp streams.
```
jonasfj: I think it's better to call them 'half-closed', as half-open is something else... sorry, I might have used the wrong terminology from time to time :)

sigurdm:
OOps. Thanks

test/server_handles_broken_connection_test.dart
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+class ClientData {
+  final int port;
+  final SendPort sendPort;
+  ClientData({this.port, this.sendPort});
+}
+
+void client(clientData) async {
```
a14n: Use `ClientData` as type for `clientData` ?

sigurdm:
Thanks for the help with this @a14n and others!

a14n:
Thank you for the fix!

PR #149, opened 2019-01-08
Minor changes to fix some major issues on mobile when using gRPC.

slightfoot:
* Fixes bug when a ResponseStream is cancelled the underlying HTTP/2 stream is not disconnected so the transport never goes inactive/idle.
* ~~Add connectTimeout to socket connections.~~
* Fixes bug where connect state is mismanaged on socket error.

PR #167, opened 2019-04-12
Close underlying connection when stream subscription is cancelled

sigurdm:
This is part of https://github.com/grpc/grpc-dart/pull/149 with an added unit test.


sigurdm:
After discussing this with @mkustermann I realized that cancelling the stream-subscription should not close the underlying http2-stream.
The reason is that the http2-stream is bidirectional. Cancelling the stream-subscription means you do not want further responses, not that you do not want to send more requests.

To have the underlying http2-stream closed the right handle to pull is `ResponseStream.cancel()`.

We could consider improving the documentation of `ResponseStream`.

I will close this for now.

PR #172, opened 2019-05-09
Add cause and trailers to GrpcError.

njovy:
Add a cause and trailers to GrpcError.

### DIFF
--- a/lib/src/client/call.dart
+++ b/lib/src/client/call.dart
@@ -63,7 +63,8 @@ class ClientCall<Q, R> implements Response {
   String get path => _method.path;
 
   void onConnectionError(error) {
-    _terminateWithError(new GrpcError.unavailable('Error connecting: $error'));
+    _terminateWithError(new GrpcError.unavailable(
+        'Error connecting: $error', error is Exception ? error : null));
   }
 
   void _terminateWithError(GrpcError error) {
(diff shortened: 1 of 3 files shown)


thelinuxfoundation:

* User @njovy isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

njovy:
I signed it

lib/src/client/call.dart
```diff
@@ -239,12 +240,15 @@ class ClientCall<Q, R> implements Response {
       // Only received a header frame and no data frames, so the header
       // should contain "trailers" as well (Trailers-Only).
       _trailers.complete(_headerMetadata);
-      final status = _headerMetadata['grpc-status'];
+      _headerMetadata.remove(":status");
+      final status = _headerMetadata.remove('grpc-status');
+//      final status = _headerMetadata['grpc-status'];
```
sigurdm: Remove out-commented code.

lib/src/client/call.dart
```diff
@@ -239,12 +240,15 @@ class ClientCall<Q, R> implements Response {
       // Only received a header frame and no data frames, so the header
       // should contain "trailers" as well (Trailers-Only).
       _trailers.complete(_headerMetadata);
-      final status = _headerMetadata['grpc-status'];
+      _headerMetadata.remove(":status");
+      final status = _headerMetadata.remove('grpc-status');
```
sigurdm: Why do you remove the header here and above?

njovy: I tried to match what grpc-java returns when it calls a trailers in the same situation. I can just return a _headerMetadata.

mit-mit:
Hi @njovy, thanks for the PR. Have you had a chance to update per the code review?

njovy:
@mit-mit I wasn't sure if I need to keep or remove values from a metadata.

PR #177, opened 2019-05-23
Don't use Transport for grpc-web

sigurdm:
Looking at the shared code for ClientConnection it doesn't really buy us anything for grpc web.
This is first step of removing the "Transport" abstraction.

Also this includes actual shutdown of open grpc-web requests when the connection is shut down (The connection maintains a list of streams).

A follow up pr will merge Http2ClientConnection with Http2Transport and remove the Transport abstraction entirely


sigurdm:
@jonasfj could you take a look?

lib/src/client/connection.dart
```diff
@@ -39,176 +39,26 @@ enum ConnectionState {
   shutdown
 }
 
-class ClientConnection {
-  final ChannelOptions options;
-  final Future<Transport> Function() _connectTransport;
+abstract class ClientConnection {
```
jonasfj: I'm starting to think this abstraction could disappear too... as there is no need to expose it from `ClientChannel`

sigurdm: Yes - I think at least from the outside.

PR #181, opened 2019-05-26
Making it possible to shutdown a client's _channel

domesticmouse:
PTAL @sigurdm 

This capability is required to port https://github.com/dart-lang/appengine to grpc.

### DIFF
--- a/lib/src/client/client.dart
+++ b/lib/src/client/client.dart
@@ -33,4 +33,8 @@ class Client {
       {CallOptions options}) {
     return _channel.createCall(method, requests, _options.mergedWith(options));
   }
+
+  Future<void> close() async {
+    await _channel.shutdown();
+  }
 }


thelinuxfoundation:

* User @domesticmouse isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

lib/src/client/client.dart
```diff
@@ -33,4 +33,8 @@ class Client {
       {CallOptions options}) {
     return _channel.createCall(method, requests, _options.mergedWith(options));
   }
+
+  Future<void> close() async {
```
sigurdm: Should this be called `shutdown()`?
Should we also have a `terminate()`?
Also this needs a doc-comment.

I guess one can construct several clients from the same channel - so when closing the client it is surprising that it closes them all. No?

Would it be better to expose the channel itself?

domesticmouse: Several clients over the same channel isn't a usecase I'd imagined. I'll reconfigure `appengine` to hold onto the client channel and shut it down as appropriate.

PR #185, opened 2019-05-28
Remove transport for http2

sigurdm:


sigurdm:
cc @jonasfj

lib/src/client/http2_connection.dart
```diff
@@ -0,0 +1,273 @@
...
+  static final _teTrailers = new Header.ascii('te', 'trailers');
+  static final _grpcAcceptEncoding =
+      new Header.ascii('grpc-accept-encoding', 'identity');
+  static final _userAgent = new Header.ascii('user-agent', 'dart-grpc/0.2.0');
+
+  final ChannelOptions options;
+
+  connection.ConnectionState _state = ConnectionState.idle;
+
+  @visibleForTesting
```
jonasfj: Is this necessary, if it's no longer part of the public interface?

The `Http2ClientConnection` class is only available if you import from `package:grpc/src/...` right?
(not that it hurts, I guess it's nice to document that this is only here for testing purposes :)

sigurdm: Good points. You are right that I made it visibleForTesting before I decided to remove the ClientConnection from the public interface. But it is strictly there for testing, so I guess I'll keep it.

PR #188, opened 2019-06-07
Export less from grpc-web.dart

sigurdm:
The ChannelOptions we have do not make sense for grpc-web. They are now http2-only.


sigurdm:
cc @jonasfj

PR #189, opened 2019-06-11
Fix const new

sigurdm:
Applies on top of: https://github.com/grpc/grpc-dart/pull/188
This is to ease merging with the main branch where this fix has been run.


sigurdm:
cc jonasfj@

sigurdm:
cc @jonasfj

PR #199, opened 2019-06-25
Fix imports

sigurdm:
Also updates the readme to announce grpc-web support.
Also updates the grpc-web example dependencies.
cc: @jonasfj

### DIFF
--- a/lib/src/auth/auth.dart
+++ b/lib/src/auth/auth.dart
@@ -16,8 +16,8 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:googleapis_auth/auth_io.dart' as auth;
-import 'package:googleapis_auth/src/crypto/rsa_sign.dart';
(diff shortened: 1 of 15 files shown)


lib/src/auth/auth.dart
```diff
@@ -16,12 +16,12 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:googleapis_auth/auth_io.dart' as auth;
-import 'package:googleapis_auth/src/crypto/rsa_sign.dart';
+import 'package:googleapis_auth/auth_browser.dart' as auth;
 import 'package:grpc/src/shared/status.dart';
```
lrhn: This import still imports from the local package using a `package:` URI. I can see you fixed all the rest :)

sigurdm: Thanks!
Fixed

Issue #196, opened 2019-06-22
Make `GrpcError` an `Exception`.


natebosch:
We recommend [not to catch `Error` instances](https://dart.dev/guides/language/effective-dart/usage#dont-explicitly-catch-error-or-types-that-implement-it).  Based on the name `GrpcError` seems like it may be an `Error`, but it's not. It's treated more like an `Exception`. Consider adding the `implements Exception` so that this is unambiguous.

Also consider changing the places that throw a String to throwing something that implements `Error`...

PR #204, opened 2019-06-26
Improve error handling

sigurdm:
Fixes: https://github.com/grpc/grpc-dart/issues/196
1) Let GrpcError implement exception

As the user is (usually) expected to catch these they should implement
Exception.
There is a bigger clean-up in splitting off the GrpcError.internal as
that seems to be for invariant violations

2) Convert some thrown strings into exceptions.

### DIFF
--- a/lib/src/client/http2_connection.dart
+++ b/lib/src/client/http2_connection.dart
@@ -71,8 +71,7 @@ class Http2ClientConnection implements connection.ClientConnection {
     var socket = await Socket.connect(host, port);
     if (_state == ConnectionState.shutdown) {
       socket.destroy();
-      // TODO(sigurdm): Throw something nicer...
-      throw 'Shutting down';
+      throw _ShutdownException();
     }
     final securityContext = credentials.securityContext;
     if (securityContext != null) {
@@ -82,8 +81,7 @@ class Http2ClientConnection implements connection.ClientConnection {
           onBadCertificate: _validateBadCertificate);
       if (_state == ConnectionState.shutdown) {
         socket.destroy();
-        // TODO(sigurdm): Throw something nicer...
-        throw 'Shutting down';
+        throw _ShutdownException();
       }
     }
     socket.done.then((_) => _handleSocketClosed());
@@ -273,3 +271,5 @@ class Http2ClientConnection implements connection.ClientConnection {
     return validator(certificate, authority);
   }
 }
+
+class _ShutdownException implements Exception {}
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -85,9 +85,7 @@ class Server {
   Service lookupService(String service) => _services[service];
 
   Future<void> serve(
-      {dynamic address,
-      int port,
-      ServerTlsCredentials security}) async {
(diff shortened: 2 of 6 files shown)


lib/src/server/server.dart
```diff
@@ -85,7 +85,10 @@ class Server {
   Service lookupService(String service) => _services[service];
 
   Future<void> serve(
-      {dynamic address, int port, ServerTlsCredentials security}) async {
+      {dynamic address,
+      int port,
+      ServerTlsCredentials security,
+      void Function(String message) logError = print}) async {
```
sigurdm: We could also use package logger. Hesitant to include another dependency...

jonasfj: Actually, it would probably be better to:
A) Use `package:logging`, or,
B) Don't print anything.

These mostly look like debug statements anyways.

sigurdm: After conferring with @lrhn I chose to swallow non-Errors and pass errors on to the zone's error handler

lib/src/server/server.dart
```diff
@@ -117,8 +122,11 @@ class Server {
         handler?.cancel();
         _connections.remove(connection);
       });
-    }, onError: (error) {
-      print('Socket error: $error');
+    }, onError: (error, stackTrace) {
+      if (error is Error) {
+        Zone.current.handleUncaughtError(error, stackTrace);
+      }
+      // Non-Errors are ignored by the server.
```
jonasfj: What exceptions could happen?

sortie: I'm not sure what could actually happen. accept(2) on Linux does have some odd errors that I'm not sure about, but most of them seem unlikely to happen for Dart because it's managed. ECONNABORTED seems super weird and unspecified when it can happen, I'd have to investigate that. The Linux manpage also seems to imply that accept(2) might receive an socket error for the accepted socket if there was one pending, also a bit surprising. Then there's the question of what errors can happen on the dart side. Nevertheless I've never seen any accept() loop actually handle such possibilities and just die when they can't accept a socket.

My gut feeling is that you should probably just die on any exception. If you find that ever happens spuriously, I'd be interested in that condition.

sigurdm:
I moved the server print changes to: https://github.com/grpc/grpc-dart/pull/212 to limit the scope of this PR.

lib/src/server/server.dart
sigurdm: https://github.com/dart-lang/sdk/blob/5721d8af6d8755fa6f70f3951de2c451c57bf12f/sdk/lib/_http/http_impl.dart#L2582 seems to ignore HandshakeException.

sigurdm:
The windows bots seems to have an unrelated problem.
Will land anyway.

lib/src/client/http2_connection.dart
```diff
@@ -275,3 +273,5 @@ class Http2ClientConnection implements connection.ClientConnection {
     return validator(certificate, authority);
   }
 }
+
+class _ShutdownException implements Exception {}
```
natebosch: This could could potentially be reasonable as an Error - from what it looks like to me it's generally indicating a programming bug?

PR #207, opened 2019-06-27
Fix interop

sigurdm:
The expression was totally broken.
Should fix (part-of) interop tests https://github.com/grpc/grpc/issues/16580

Side-note: the lack of complaints about this probably means we have no active users of jwt auhtorization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.0.0+2
+
+* Fix computation of authority. This should fix authorization.
+
 ## 2.0.0+1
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.0+2
+
+* Fix computation of authority. This should fix authorization.
```
jonasfj: Please bump the patch version... Build metadata is usually only for repacking or typos in comments, readme, pubspec...

https://semver.org/#spec-item-10

jonasfj: This actually fixes something :)

sigurdm: You are right! Done!

PR #212, opened 2019-07-02
Don't print from server

sigurdm:


lib/src/server/server.dart
```diff
@@ -107,8 +107,11 @@ class Server {
       // timeout.
       connection.incomingStreams.listen((stream) {
         handler = serveStream_(stream);
-      }, onError: (error) {
-        print('Connection error: $error');
+      }, onError: (error, stackTrace) {
+        if (error is Error) {
+          Zone.current.handleUncaughtError(error, stackTrace);
+        }
+        // Non-Errors are ignored by the server.
```
jonasfj: I think you should consider only handling the handshake exception, as this is what  `HttpServer` does... and a grpc server is really just an HTTP server :)

But I also don't think we're likely to see other exceptions, so it probably doesn't matter much.

mit-mit:
@mraleph looks like this can be merged?

Issue #228, opened 2019-08-12
Not properly handling http/2 connection becoming not healthy


tvolkert:
Spawned from https://github.com/dart-lang/appengine/issues/117 and https://github.com/dart-lang/http2/issues/46.

For consolidation, I'll leave the original reported issue text the same below and copy any relevant comments over from the other issues.

---

`package:appengine` registers a single `gcloud:DatastoreDB` instance upon server startup, which triggers the following objects to be persisted as part of its object graph:
*  ↳ gcloud:DatastoreDB maintains a reference to a `gcloud:Datastore`
    * (implemented by `appengine:GrpcDatastoreImpl`)
*  ↳ `appengine:GrpcDatastoreImpl` maintains a reference to `grpc:ClientChannel`
*  ↳ `grpc:ClientChannel` maintains a reference to `grpc:Http2ClientConnection`
*  ↳ `grpc:Http2ClientConnection` maintains a reference to `http2:ClientTransportConnection`

I'm seeing cases where the following code is hitting [these lines](https://github.com/dart-lang/http2/blob/04843721d4778dcba8e9b490efca9c0ea809fec8/lib/src/connection.dart#L354-L356):

```dart
} else if (frame is GoawayFrame) { 
  _streams.processGoawayFrame(frame); 
  _finishing(active: false);
}
```

This causes the connection's state to get marked as `IsFinishing/PassiveFinishing`. Then later, grpc tries to issue another request [here](https://github.com/grpc/grpc-dart/blob/1213bc546b46da385afc94e645dd44fefcfb7f82/lib/src/client/http2_connection.dart#L130):

```dart
GrpcTransportStream makeRequest(String path, Duration timeout,
    Map<String, String> metadata, ErrorHandler onRequestFailure) {
  final headers = createCallHeaders(
      credentials.isSecure, authority, path, timeout, metadata,
      userAgent: options.userAgent);
  final stream = _transportConnection.makeRequest(headers);
  return Http2TransportStream(stream, onRequestFailure);
}
```

And it results in the [following error](https://github.com/dart-lang/http2/blob/04843721d4778dcba8e9b490efca9c0ea809fec8/lib/src/connection.dart#L461-L463):

```dart
throw new StateError(
    'The http/2 connection is finishing and can therefore not be used to '
    'make new streams.');
```

[...]

PR #231, opened 2019-08-16
Improve connection handling

sigurdm:
Most of this is appropriated from the grpc implementation we had in https://github.com/dart-lang/appengine. Big thanks to @mkustermann for exlpaining the intricate details!

This should make the connection handling a lot more robust, and also faster.
Hopefully dealing with: https://github.com/grpc/grpc-dart/issues/228


lib/src/client/http2_connection.dart
```diff
@@ -69,25 +73,36 @@ class Http2ClientConnection implements connection.ClientConnection {
...
-    }
     final securityContext = credentials.securityContext;
-    if (securityContext != null) {
-      socket = await SecureSocket.secure(socket,
-          host: authority,
+    Socket socket;
+    if (securityContext == null) {
+      socket = await Socket.connect(host, port);
+    } else {
+      final socket = await SecureSocket.connect(host, port,
```
tvolkert: You're creating a new local variable here, so the outer `socket` remains null.  Perhaps name this `secureSocket`, and then set `socket = secureSocket` after the "h2" if check below?

tvolkert: (also, is there any way to test this code path?)

lib/src/client/http2_connection.dart
```diff
@@ -69,25 +73,36 @@ class Http2ClientConnection implements connection.ClientConnection {
...
       }
     }
-    socket.done.then((_) => _handleSocketClosed());
-    return ClientTransportConnection.viaSocket(socket);
+
+    final connection = ClientTransportConnection.viaSocket(socket);
+    socket.done.then((_) => _abandonConnection());
+
+    // Give the settings settings-frame a bit of time to arrive.
+    await new Future.delayed(_estimatedRoundTripTime);
```
tvolkert: This seems like a smell - can we not be notified somehow when we're ready?

lib/src/client/http2_connection.dart
```diff
@@ -96,17 +111,45 @@ class Http2ClientConnection implements connection.ClientConnection {
...
+  /// Abandons the current connection if it is unhealthy or has been open for
+  /// too long.
+  ///
+  /// Assumes [_transportConnection] is not `null`.
+  void _refreshConnectionIfUnhealthy() {
+    final bool isHealthy = _transportConnection.isOpen;
+    final bool shouldRefresh =
+        _connectionLifeTimer.elapsed > options.connectionTimeout;
+    if (!isHealthy || shouldRefresh) {
+      if (isHealthy) {
```
tvolkert: This is the same as `if (shouldRefresh)`, and there's no reason for it to be in the outer `if` block.

tvolkert:
The old `appengine` code used a Dialer to ensure only one new http/2 connection is made if multiple event loops try to establish a new connection.  Does this handle that case?

lib/src/client/http2_connection.dart
sigurdm: Oops
Thanks for catching this!

lib/src/client/http2_connection.dart
jonasfj: Maybe add a comment that explains why we do this, and how it could be better.

tvolkert:
Thanks @sigurdm - can you publish this to Pub too?

sigurdm:
Yep - done

PR #233, opened 2019-08-22
Fix use of authority when making a secure connection

sigurdm:
This should fix https://github.com/grpc/grpc/issues/20017

### DIFF
--- a/lib/src/client/http2_connection.dart
+++ b/lib/src/client/http2_connection.dart
@@ -77,19 +77,13 @@ class Http2ClientConnection implements connection.ClientConnection {
 
   Future<ClientTransportConnection> connectTransport() async {
     final securityContext = credentials.securityContext;
-    Socket socket;
-    if (securityContext == null) {
-      socket = await Socket.connect(host, port);
-    } else {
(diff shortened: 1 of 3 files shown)


lib/src/client/http2_connection.dart
```diff
@@ -77,19 +77,13 @@ class Http2ClientConnection implements connection.ClientConnection {
...
-    if (securityContext == null) {
-      socket = await Socket.connect(host, port);
-    } else {
-      socket = await SecureSocket.connect(host, port,
-          supportedProtocols: ['h2'],
+    Socket socket = await Socket.connect(host, port);
+    if (securityContext != null) {
+      // Todo(sigurdm): We want to pass supportedProtocols: ['h2']. http://dartbug.com/37950
+      socket = await SecureSocket.secure(socket,
+          host: authority,
```
jonasfj: If `host == authority` as is the common case, you could still use `connect`. But this seems like an over optimization :)

sigurdm: Yes - I thought it would be even more cases for our undertested framework

I hope we solve http://dartbug.com/37950 and move to a better world.

Issue #219, opened 2019-07-26
ChannelCredentials.secure


qiuguian:
<short description>

<version of the grpc-dart packages used; see your `pubspec.lock` file>

## Repro steps

1. step1
1. step2
1. step3

Expected result: <describe what should have happened>

Actual result: <describe what actually happened>

## Details

<Include any other relevant details, logs, etc.>

friddle:
你应该问的更具体+最好用英文问。

qiuguian:
run in ios is no error but  run in android use flutter,thent got error:Error connecting: HandshakeException: Handshake error in client (OS Error: 
I/flutter ( 8621):   CERTIFICATE_VERIFY_FAILED: self signed certificate(handshake.cc:352))

friddle:
你的地址假如和你证书的地址对不上。
即证书的公用名必须和你的请求地址对上。

或者你需要加上secure的authority
```
    return ClientChannel('appapi.github.com',
        port: 443,
        options:
        ChannelOptions(credentials: ChannelCredentials.secure(
            certificates: sslKey, authority: "appapi.github.com"
        )));
```

friddle:
还有一个问题开多个issues一样没人看的。。。

sigurdm:
I am not sure I follow this issue very well. I don't read Chinese.
Any chance you could clarify the issue description in English?

But it is surprising if there is a different behavior between ios and android.
@sortie do you know if there are differences between the Android and ios embeddings that could cause the HandshakeException?

friddle:
@sigurdm  seem's like the problem is he forget to input  ChannelCredentials.secure authority options

sigurdm:
Is this solved?

sortie:
I'm unsubscribing from this issue since the notifications go to my personal mail as this repository is not inside the google organization. If you have any action items for me, please file a Dart SDK issue.

nevi-me:
I also can't read Chinese, but I'm wondering if the issue is the same as the one I've started noticing since upgrading from a very old grpc version (0.x) to the latest.

@sigurdm it looks like the `secure` channel's broken on the latest version. When I try to connect to a grpc server using the default `ChannesCredentials.secure()` option, I'm unable to connect.

Passing the `authority` on `secure()` above fixed it. If it's intended behaviour that the authority always be present, we might need to document it somewhere. Where can I do that, if it's the case and hasn't been documented? Thanks

sigurdm:
[...]

Issue #237, opened 2019-09-04
Certificate validation fails when trying to establish secure connection to endpoint on port other than 443


tp:
As I commented https://github.com/grpc/grpc-dart/issues/50#issuecomment-527879094 I have some issue connecting to a server with a self-signed certificate on port 50051 on localhost, while it works fine on our staging setup on port 443.

> When trying to access it locally (localhost:50051) I get an error:

```
gRPC Error (14, Error connecting: HandshakeException: Handshake error in client (OS Error: 
        CERTIFICATE_VERIFY_FAILED: ok(handshake.cc:352)))
```

~~After some debugging I think that it's related to the authority building added in https://github.com/grpc/grpc-dart/pull/207~~ (see below, I think the authority is not the real breaking change here, but it's usage in establishing the secure connection)

This is an excerpt from my certificate:

```
…
            X509v3 Subject Alternative Name: 
                DNS:localhost, DNS:grips.staging.de, DNS:grips.production.de, IP Address:127.0.0.1, IP Address:0.0.0.0
    Signature Algorithm: sha256WithRSAEncryption
…
```

So I guess in order to comply with the authority in [here](https://github.com/sigurdm/grpc-dart/blob/35d627dbb248ce1e846d992bcd75e6048861b0ff/lib/src/client/http2_connection.dart#L65)  I would have to add an alternative name like `localhost:50051` to the certificate.

My question though then is whether it's correct to specify ports in the certificate (some quick googling indicated it's not a common/recommended practice), or whether we should change the authority calculation?

tp:
Ok, seems like for the authority that is desired to include the port
https://github.com/grpc/grpc-dart/commit/b7be3af34bba6a6f2ea2862e8d5397cbcda59abc / https://stackoverflow.com/questions/50034792/should-the-http-2-authority-header-include-port-number

But then there is the question why this changed from passing the `host` to passing the `authority` to create the connection.

cc @sigurdm as you were driving that linked change.

tp:
I had the error with version `2.0.0+1`, so I thought above trying the latest versions (even thought I initially assumed security wouldn't be prone to much change, but since this is under active development why not try it out.)

It's now fine on version 2.1.0 (which uses `host`) as well as 2.1.1 (which uses `authority`).

So it seems solved to me. There's probably some other change in the 2.0.0 to 2.1.0 diff that affected how the certificate validation works which I have not found yet.

tp:
[...]

PR #238, opened 2019-09-05
Use host without port as authority when making secure connection

sigurdm:
Should fix https://github.com/grpc/grpc-dart/issues/219 and https://github.com/grpc/grpc-dart/issues/237

### DIFF
--- a/lib/src/client/http2_connection.dart
+++ b/lib/src/client/http2_connection.dart
@@ -81,7 +81,11 @@ class Http2ClientConnection implements connection.ClientConnection {
     if (securityContext != null) {
       // Todo(sigurdm): We want to pass supportedProtocols: ['h2']. http://dartbug.com/37950
       socket = await SecureSocket.secure(socket,
-          host: authority,
+          // This is not really the host, but the authority to verify the TLC
+          // connection against.
+          //
+          // We don't use `this.authority` here, as that includes the port.
+          host: options.credentials.authority ?? host,
           context: securityContext,
           onBadCertificate: _validateBadCertificate);
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.1.2
+
+* Fix bug introduced in 2.1.1 where the port would be added to the default authority when making a
+  secure connection.
+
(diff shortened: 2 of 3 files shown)


lib/src/client/http2_connection.dart
```diff
@@ -81,7 +81,7 @@ class Http2ClientConnection implements connection.ClientConnection {
     if (securityContext != null) {
       // Todo(sigurdm): We want to pass supportedProtocols: ['h2']. http://dartbug.com/37950
       socket = await SecureSocket.secure(socket,
-          host: authority,
+          host: options.credentials.authority ?? host,
```
tp: @sigurdm Since there was all this confusion around the HTTP2 authority and "this" authority (which as far as I can tell is always without a port and just a way to allow certs from a different host), should there maybe be a comment on `authority` to disambiguate it from the other `authority` field?


Also is maybe the `host` parameter name on `SecureSocket.secure` misleading and should rather the `authority`? Because the actual connection to the `host` already gets made in line 80 (`Socket.connect(host, port);`).

sigurdm: Good idea to leave a comment.

I don't think we can change the argument name on `SecureSocket` easily.

jonasfj: It might be nice with a comment here explaining that the `host` is actually the `authority` we're verifying the TLS connection to belong to.

sigurdm: Done

sigurdm:
Hmm - not sure why the cla/linuxfoundation check doesn't go through.
@mit-mit do you know anything?

mit-mit:
I don't, but perhaps @jtattermusch does?

jonasfj:
Maybe it got stuck... try to close and reopen, or file as new PR.

jtattermusch:
Ad CLA check:
I think the Required statuses are configured in the repo admin section (but I myself don't have access to grpc-dart repo config) - I think instead of cla/linuxfoundation the CLA check is now "EasyCLA".
Alternatively, I'd try force-pushing after rebase to get a different commit SHA.

PR #240, opened 2019-09-19
make accessToken in BaseAuthenticator accessible

Ragnaroek:
Use-Case is an own implementation of Token based access control that wants to re-use the BaseAuthenticator functionality.

### DIFF
--- a/lib/src/auth/auth.dart
+++ b/lib/src/auth/auth.dart
@@ -26,19 +26,19 @@ import 'rsa.dart';
 const _tokenExpirationThreshold = Duration(seconds: 30);
 
 abstract class BaseAuthenticator {
-  auth.AccessToken _accessToken;
+  auth.AccessToken accessToken;
   String _lastUri;
 
   Future<void> authenticate(Map<String, String> metadata, String uri) async {
     if (uri == null) {
       throw GrpcError.unauthenticated('Credentials require secure transport.');
     }
-    if (_accessToken == null || _accessToken.hasExpired || uri != _lastUri) {
+    if (accessToken == null || accessToken.hasExpired || uri != _lastUri) {
       await obtainAccessCredentials(uri);
       _lastUri = uri;
     }
 
-    final auth = '${_accessToken.type} ${_accessToken.data}';
+    final auth = '${accessToken.type} ${accessToken.data}';
     metadata['authorization'] = auth;
 
     if (_tokenExpiresSoon) {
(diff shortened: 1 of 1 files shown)


Ragnaroek:
@googlebot I signed it!

Ragnaroek:
@googlebot I signed it!

sigurdm:
I am not against this.

I don't like the design of the class hierarchy that much, thinking that the BaseAuthenticator should probably not have been exposed like that, but that's what we have for now.

@jonasfj what do you think?

jonasfj:
I would either:

-----
A) Create a new base class for `BaseAuthenticator` called `Authenticator` and replace uses of `BaseAuthenticator` throught-out the code-base... (also fix the documentation todos).

```dart
// TODO: Document this
abstract class Authenticator {
  // TODO: Document this
  Future<void> authenticate(Map<String, String> metadata, String uri);
  // TODO: Document this
  CallOptions get toCallOptions => CallOptions(providers: [authenticate]);
}

@deprecated
abstract class BaseAuthenticator extends Authenticator {
  // remains as is...
}
```
Then follow up with something like:
```dart
// TODO: Document this
abstract class Authenticator {
  // TODO: Document this
  Future<void> authenticate(Map<String, String> metadata, String uri);
  // TODO: Document this
  CallOptions get toCallOptions => CallOptions(providers: [authenticate]);

  /// Create an [Authenticator] given a method [getAccessToken] which
  /// can obtain an [auth.AccessToken] for a given URI.
  static Authenticator create(Future<auth.AccessToken> Function(String uri) getAccessToken)
    => _Authenticator(getAccessToken);
}

@deprecated
abstract class BaseAuthenticator extends Authenticator {
  // remains as is...
}

class _Authenticator extends BaseAuthenticator {
  final Future<auth.AccessToken> Function(String uri) _getAccessToken;

  _Authenticator(this._getAccessToken);

  @override
  Future<void> obtainAccessCredentials(String uri) async
    => _accessToken = await _getAccessToken(uri);
}
```


------------------
B) Create a static `BaseAuthenticator.create` method that accepts a function to do this, something like this:

```dart
abstract class BaseAuthenticator {
  // remains as is...

  /// Create a [BaseAuthenticator] given a method [getAccessToken] which
  /// can obtain an [auth.AccessToken] for a given URI.
  static BaseAuthenticator create(Future<auth.AccessToken> Function(String uri) getAccessToken)
    => _Authenticator(getAccessToken);
}

class _Authenticator extends BaseAuthenticator {
  final Future<auth.AccessToken> Function(String uri) _getAccessToken;

  _Authenticator(this._getAccessToken);

  @override
  Future<void> obtainAccessCredentials(String uri) async
    => _accessToken = await _getAccessToken(uri);
}
```

It's less clean-up than solution (A), but also a lot easier to implement.

Note. I would strongly advice a method like `BaseAuthenticator.create` be documented with an inline-example in addition to some explanation.


-----
Warnings, this just quick code snippets, I haven't tested or fully thought this through.

mit-mit:
Closing due to unresolved design issues.

PR #242, opened 2019-09-27
Add explicit 'localhost' to tests involving server

nichite:
This was breaking internal tests when not specified. Also, remove an unnecessary import, and run dartfmt on the whole repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 ## 2.1.3
 
 * Fix: explicitly specify `'localhost'` for round trip tests
-* Remove unnecessary import of HTTP2 lib
-* Run `dartfmt`
 
 ## 2.1.2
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.1.3
+
+* Fix: explicitly specify `'localhost'` for round trip tests
+* Remove unnecessary import of HTTP2 lib
+* Run `dartfmt`
```
sigurdm: Keep the Changelog to user-relevant changes (IE. only the first item)

nichite: Done.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.3
+
+* Fix: explicitly specify `'localhost'` for round trip tests
```
sigurdm: Reading this again, this is just a test.
I think we can skip publishing this in a version for itself.

nichite: Sure, reverted.

PR #252, opened 2019-11-28
added logging

DisDis:


DisDis:
@sigurdm ping

sigurdm:
@jonasfj we spoke about doing something like this before, i don't remember our conclusion.

jonasfj:
Uh, I don't recall either. Adding dependency on `package:logging` doesn't seem crazy, I don't know about all the users.

But should we printing these things at all?

sigurdm:
That was probably the question where we stranded last time.
We don't really know what causes these errors...

DisDis:
@sigurdm @jonasfj If we use 'print' then user can not manage format of log. It is critical for log aggregation system, k8s and e.t.c. If we use 'package:logging' then user can manager all log records.

DisDis:
@sigurdm @jonasfj ping

jonasfj:
@DisDis, I think a question here is:
 * What are the possible errors/exceptions here?
 * Is it correct to ignore them, or is it better to crash? (if we ignore them and the server simply goes unhealthy, it's better crash).

If it's correct to log and ignore the errors, then this is a solid PR. If it's not correct, then adding dependency upon `package:logging` is unnecessary and potentially disruptive to downstream consumers of this package.

DisDis:
@jonasfj ok, i close PR

PR #253, opened 2019-11-28
Share port support

DisDis:


DisDis:
@sigurdm ping

lib/src/server/server.dart
```diff
@@ -99,7 +102,7 @@ class Server {
       server = _secureServer;
```
sigurdm: You need to also pass the parameters in the secure case

DisDis: done

DisDis:
@sigurdm please merge this PR

DisDis:
@sigurdm ping

PR #254, opened 2019-11-28
Isolate service example

DisDis:
after 
#251
#252
#253


DisDis:
I don't understand how to do the correct share port test. Any ideas?

Issue #121, opened 2018-09-08
Feature Request: Client Interceptors


nevi-me:
It looks like the implemented interceptors are only server interceptors. Is it possible to create client interceptors?

I need to intercept calls that I make from a Flutter application and add authentication metadata

Using v 0.6.4

## Repro steps

None

## Details

I don't think this would need a gRFC, as I presume a general one exists since Java has client interceptors
.

sigurdm:
I am not against client interceptors, but it is not on the current list of priorities.

We are open for contributions...

TinoGuo:
I found a solution but not verify.

In j
```
class ExampleClientChannel extends ClientChannel {
  ExampleClientChannel(String host, {int port, ChannelOptions options})
      : super(host, port: port, options: options);

  ClientConnection _connection;

  bool _isShutdown = false;

  /// Shuts down this channel.
  ///
  /// No further RPCs can be made on this channel. RPCs already in progress will
  /// be allowed to complete.
  Future<Null> shutdown() {
    if (_isShutdown) return new Future.value();
    _isShutdown = true;
    return _connection.shutdown();
  }

  /// Terminates this channel.
  ///
  /// RPCs already in progress will be terminated. No further RPCs can be made
  /// on this channel.
  Future<Null> terminate() {
    _isShutdown = true;
    return _connection.terminate();
  }

  /// Returns a connection to this [Channel]'s RPC endpoint.
  ///
  /// The connection may be shared between multiple RPCs.
  Future<ClientConnection> getConnection() async {
    if (_isShutdown) throw new GrpcError.unavailable('Channel shutting down.');
    return _connection ??= new ArkClientCall(host, port, options);
  }

  /// Initiates a new RPC on this connection.
  ClientCall<Q, R> createCall<Q, R>(
      ClientMethod<Q, R> method, Stream<Q> requests, CallOptions options) {
    final call = new ClientCall(method, requests, options);
    getConnection().then((connection) {
      if (call.isCancelled) return;
      connection.dispatchCall(call);
    }, onError: call.onConnectionError);
    return call;
  }
}
```

TinoGuo:
I found a solution but not verify.

Inherit from ClinetChannel, and just copy the ClientChannel source code.
```
class ExampleClientChannel extends ClientChannel {
  ExampleClientChannel(String host, {int port, ChannelOptions options})
      : super(host, port: port, options: options);

  ClientConnection _connection;

  bool _isShutdown = false;

[...]

PR #261, opened 2019-12-20
Implement client-side interceptor

lqs:
This implements client-side interceptor (also mentioned in #121), which is helpful for tracking request duration or adding global error handler.

Usage example:
```
ClientInterceptor _interceptor = (method, requests, options, invoker) {
  final startTime = DateTime.now();
  print("start call ${method.path}");

  final call = invoker(method, requests.map((request) {
    print("request: ${request.runtimeType}($request)");
    return request;
  }), options);

  call.headers.then((headers) {
    print("headers: $headers");
  });
  call.trailers.then((trailers) {
    print("trailers: $trailers");
    final endTime = DateTime.now();
    print(endTime.difference(startTime));
  });

  call.response.listen((response) {
    print("response value: ${response.runtimeType}($response)");
  });

  return call;
};

var channel = ClientChannel("host-of-grpc-service", interceptor: _interceptor);
```

---

Admin-comment: Fixes https://github.com/grpc/grpc-dart/issues/121


cosmotek:
does this provide access to the request metadata? I think something that needs to be possible here is changing metadata (auth) on specific error and re-execing the call.

cosmotek:
I've been playing with implementing the same thing for the last hour or so.... Not sure if I know what I am doing, but I realize that these changes require changes to the protoc-gen-dart generator project as well.

lqs:
@cosmotek Yes, the request metadata can be accessed in `options.metadata`

phamnhuvu-dev:
This code doesn't support `async function`. Sometimes you will need call `async function` to get data for `metadata`
I wrote an interceptor for `async function` to update `options.metadata` but now I am busy with the deadline. Sorry 
See you later soon.

upday7:
this this pr being reviewed? .. hope this will be in next release ..

gustav3d:
Hows review going ?

cq-z:
@lqs support web code
web_channel.dart:
``GrpcWebClientChannel.xhr(this.uri,interceptor)` : super(interceptor: interceptor);
``

nateshmbhat:
any progress on this ?

nateshmbhat:
@lqs how do we modify the response here ?

Only the request object passes through the interceptor. So modification to the request is possible.

How about modification to the response ? here we are only listening to the response. any idea how to add that feature to the interceptor. ?

Arpit1496:
Why this is not merged yet ?

Arpit1496:
Why this is not merged yet ?

mraleph:
We are reviewing a more feature complete version of this functionality here https://github.com/grpc/grpc-dart/pull/338. I am going to close this PR in favor of #338.

PR #269, opened 2020-01-19
Update protobuf to ^1.0.0

firephreek:
Set protobuf version to ^1.0.0 per https://github.com/dart-lang/protobuf/issues/310
Without this change, the tutorial/walkthrough is effectively broken which creates confusion for new users.


sigurdm:
Thanks for the contribution!

This would also require regenerating the pb.dart files with newest protoc.

sigurdm:
This looks good. We need the CLA signed before landing

PR #272, opened 2020-01-22
Pass service to interceptors

jey:
A single `grpc.Server` can have multiple services, so interceptors may need to know which service they're acting on. Just having the method and its name is insufficient.

### DIFF
--- a/lib/src/server/interceptor.dart
+++ b/lib/src/server/interceptor.dart
@@ -11,4 +11,4 @@ import 'service.dart';
 /// If the interceptor throws [Exception], [GrpcError.internal] with exception.toString() will be returned.
 /// If the interceptor returns null, the corresponding [ServiceMethod] of [Service] will be called.
 typedef Interceptor = FutureOr<GrpcError> Function(
-    ServiceCall call, ServiceMethod method);
+    ServiceCall call, Service service, ServiceMethod method);
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,7 +1,7 @@
 name: grpc
 description: Dart implementation of gRPC, a high performance, open-source universal RPC framework.
 
-version: 2.1.3
+version: 2.2.0
 
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/grpc-dart
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -131,7 +131,7 @@ class ServerHandler_ extends ServiceCall {
   Future<GrpcError> _applyInterceptors() async {
     try {
       for (final interceptor in _interceptors) {
-        final error = await interceptor(this, this._descriptor);
+        final error = await interceptor(this, _service, _descriptor);
         if (error != null) {
           return error;
         }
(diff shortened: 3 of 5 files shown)


jey:
Actually, shouldn't the interceptor be responsible for invoking the underlying method too? That's how it works go-grpc's [UnaryServerInterceptor](https://godoc.org/google.golang.org/grpc#UnaryServerInterceptor).

This would allow many additional uses of interceptors: method timing, error logging, etc

pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: grpc
 description: Dart implementation of gRPC, a high performance, open-source universal RPC framework.
 
-version: 2.1.3
```
mraleph, line 4: This would need a CHANGELOG.md entry describing the change.

lib/src/server/interceptor.dart
```diff
@@ -11,4 +11,4 @@ import 'service.dart';
 /// If the interceptor throws [Exception], [GrpcError.internal] with exception.toString() will be returned.
 /// If the interceptor returns null, the corresponding [ServiceMethod] of [Service] will be called.
 typedef Interceptor = FutureOr<GrpcError> Function(
-    ServiceCall call, ServiceMethod method);
+    ServiceCall call, Service service, ServiceMethod method);
```
mraleph, line 14: What about just exposing this information through `ServiceCall`, e.g. you can add

```dart
class ServiceCall {
  /// ...
  Service get service;
  
  /// ...
  ServiceMethod get method;
}
```

Then you can access this information through `call` and you would not need to make a breaking change to `Interceptor` signature.

mraleph (changes requested):
Thanks for your contribution and apologies for the delay in review.

See suggestions below. Additionally this needs at least some tests which access `service` and verify that it's correct.

mraleph:
> Actually, shouldn't the interceptor be responsible for invoking the underlying method too? That's how it works go-grpc's [UnaryServerInterceptor](https://godoc.org/google.golang.org/grpc#UnaryServerInterceptor).
> 
> This would allow many additional uses of interceptors: method timing, error logging, etc

Yeah, I noticed the discrepancy between how interceptors work in Dart and other implementations. 

If you are still interested in refactoring that feel free to submit a separate PR for that.

shaxbee:
> > Actually, shouldn't the interceptor be responsible for invoking the underlying method too? That's how it works go-grpc's [UnaryServerInterceptor](https://godoc.org/google.golang.org/grpc#UnaryServerInterc
> If you are still interested in refactoring that feel free to submit a separate PR for that.

@mraleph I'm happy to refractor using similar interface to ClientInterceptor.

mraleph:
> @mraleph I'm happy to refractor using similar interface to `ClientInterceptor`.

This would be breaking change - so there should be either strong demand for it or it should be clear that not so many users even use interceptors. It is unclear if either is true for interceptors. 

If you would like to submit refactoring maybe we could first _add_ another interceptor mechanism which is more aligned with other implementations and mark current interceptor mechanism deprecated. Then we can remove current interceptor mechanism at a later date after few releases.

mraleph:
@jey if you are still interested in landing this PR let me know. For now I am going to close it as stale.

PR #274, opened 2020-01-24
server: Cancel call when the outgoing messages are closed

sigurdm:
This should fix: https://github.com/grpc/grpc/issues/21755


jtattermusch:
Thanks @sigurdm for looking into this. Can we merge this now to fix the interop test flake?

sigurdm:
Yes - I hoped to write a test, but that can come later.

jtattermusch:
Ok - let's merge then - looks like I don't have the merge rights on this repo.

PR #292, opened 2020-04-14
Export more symbols for alternate implementation of `ClientChannel`

jiahaog:
To create an alternate implementation of `ClientChannel`, we currently need to import symbols from `lib/src` which are "[private](https://dart.dev/tools/pub/package-layout#implementation-files)" to the package. This change exports them as part of the public API.

Internal bug: b/152361560

### DIFF
--- a/lib/grpc_connection_interface.dart
+++ b/lib/grpc_connection_interface.dart
@@ -1,0 +1,30 @@
+// Copyright (c) 2020, the gRPC project authors. Please see the AUTHORS file
+// for details. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


jiahaog:
@nichite please take a look!

jiahaog:
Thanks! Most of the exposed symbols are just opaque symbols, and I don't think it's a good idea to duplicate over helper logic like [`frame`](https://github.com/grpc/grpc-dart/blob/03ecb840641f1f9cf99f63b1dae042dd40f1b209/lib/src/shared/message.dart#L57), in case it changes in the future with compression. Not sure if it will still be compatible, but we can revisit that when the time comes :)

PR #293, opened 2020-04-23
Make sure connection errors are propagated

robsonmeemo:


lib/src/shared/message.dart
```diff
@@ -69,8 +70,8 @@ StreamTransformer<GrpcMessage, GrpcMessage> grpcDecompressor() =>
         handleData: (GrpcMessage value, EventSink<GrpcMessage> sink) {
       if (value is GrpcData) {
         if (value.isCompressed) {
```
sigurdm, line 72: This is cool, but will not work on for grpc_web. That doesn't have access to dart:io.

Issue #249, opened 2019-11-21
Better performance for grpc-dart


jquenel:
gRPC clients and servers are very slow.

grpc-dart vesion 2.1.3
Ubuntu 19.04

## Repro steps

1. Open the helloworld example
2. Add some lines around the client rpc call to time it, and optionally repeat it.

```dart
while (true) {
  try {
      var t1 = DateTime.now().millisecondsSinceEpoch;
      final response = await stub.sayHello(HelloRequest()..name = name);
      var t2 = DateTime.now().millisecondsSinceEpoch;

      print('Greeter client received: ${response.message}');
      print((t2 - t1).toString());
    } catch (e) {
      print('Caught error: $e');
    }
  }
```

3. Launch client and server on same machine.

Expected result:  high performance rpc

Actual result: a minimum of 80ms for each call

## Details

I don't know if that is relevant, but the transfer speed seems to be limited too. I tried running a c++ server with a Dart client requesting large files, and had transfer speeds that would not go beyond 10 Mb/s. On the same machine, a c++ client making the same requests has transfer speeds of 400 Mb/s.

Am I doing something wrong with the client ?

soach:
i did a test to repeat the problem by running the samplest example:helloworld, using  go ,java ,and dart language.
when a java grpc server is runing , go and java client work fine with less than 1ms per rpc call. but dart grpc client needs 40ms.
when a dart server is runing , go and java grpc client need 40ms, but dart client needs 80ms.
that seems both dart client and dart server has some thing wong ? 
Am I doing something wrong with the client ?

jquenel:
Any idea on how we could help track the issue or improve this ? I would really love to see dart and grpc be a competitive solution in my Flutter app, and not rely on some dart.ffi shenanigans.

soach:
i found the root cause. 
the latest version of grpc dart doest not turn on tcp no delay. So the solution is turn it on .
socket.setOption(SocketOption.tcpNoDelay, true);

soach:
> Any idea on how we could help track the issue or improve this ? I would really love to see dart and grpc be a competitive solution in my Flutter app, and not rely on some dart.ffi shenanigans.

the fixed code is :
server.dart
server.listen((socket) {
      **socket.setOption(SocketOption.tcpNoDelay, true);**
      final connection = ServerTransportConnection.viaSocket(socket,
          settings: http2ServerSettings);
      _connections.add(connection);

and client:
http2_connection.dart:
[...]

PR #275, opened 2020-01-28
fix latency problem when using it as a grpc server or client.

soach:
fix the issue : https://github.com/grpc/grpc-dart/issues/249

PR #298, opened 2020-05-11
Use tcpNoDelay option for sockets

sigurdm:
This supersedes/replicates https://github.com/grpc/grpc-dart/pull/275.

Fixes https://github.com/grpc/grpc-dart/issues/249


gustav3d:
Lets hope there is more activity on this then 275.
Or if the devs are not active anymore, can you please just give commit rights to anybody ?!  like sigurdm. So the project can move forward.

PR #300, opened 2020-05-13
Remove 20ms latency when connect

hubaxis:
We are using grpc for rpc between microservices. Usually it ~1k requests per second, 20ms delay when connecting incredible long for us

### DIFF
--- a/lib/src/client/http2_connection.dart
+++ b/lib/src/client/http2_connection.dart
@@ -73,8 +73,6 @@ class Http2ClientConnection implements connection.ClientConnection {
 
   ConnectionState get state => _state;
 
-  static const _estimatedRoundTripTime = const Duration(milliseconds: 20);
-
   Future<ClientTransportConnection> connectTransport() async {
     final securityContext = credentials.securityContext;
     Socket socket = await Socket.connect(host, port);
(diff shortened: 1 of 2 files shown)


hubaxis:
Any progress here?

gustav3d:
Hardly.
The people with commit rights are inactive, preventing any progress of the
project.

mit-mit:
It seems dangerous to just remove this; there is a TODO stating that we'd need a new API from HTTP2.

@hubaxis have you been running with this patch for a while succesfully?

mraleph:
We can't unfortunately completely remove this code - I have discussed this with @mkustermann who explained that there were real world reasons to put this sort of hack: a situation might arise where server concurrency level is lower than what you expect by default, so if you hurry too much you will receive errors back. Waiting for the first settings frame addresses this issue at least partially (server alter settings at any time by sending a new SETTINGS frame - so there is some amount of racing built into the design here).

Specification allows the client to send frames before it receives the first settings frame, however it also requires clients to honor server settings:

> To avoid unnecessary latency, clients are permitted to send
additional frames to the server immediately after sending the client
connection preface, without waiting to receive the server connection
preface.  It is important to note, however, that the server
connection preface SETTINGS frame might include parameters that
necessarily alter how a client is expected to communicate with the
server.  Upon receiving the SETTINGS frame, the client is expected to
honor any parameters established.  In some configurations, it is
possible for the server to transmit SETTINGS before the client sends
additional frames, providing an opportunity to avoid this issue.

Here is what I suggest: 

- I have started a PR for `http2` package to expose a future which fires when connection becomes operational (after receiving the first SETTINGS frame). This would allow us to replace this 20ms delay hack with 
    ```dart
    await connection.onConnectionOperational;
    ``` 
- We add `ChannelOptions.waitForOperational` which is set to `true` by default (safe default) which users can switch to `false` if they know what they are doing.

How does that sound @hubaxis?

mit-mit:
@hubaxis are you still interested in working on this PR?

mraleph:
Closing as assumed stale for now.

PR #302, opened 2020-05-13
Close transport connection properly

hubaxis:
Sometimes connection never finished. Force terminate it after timeout


hubaxis:
Any progress here?

long1eu:
I someone watching this PR's?

mit-mit:
This looks like a good, small change, cc @mraleph

mraleph (changes requested):
- Please describe what we want to achieve by terminating connection forcefully, e.g. we want to prevent resource leak when connection is just hanging out there.
- If possible provide test case, if not at least describe why you want this change to be made, how did you discover that this happens and is an issue.

mraleph:
Closing assuming stale.

PR #303, opened 2020-05-13
Not to fail pending requests, just reschedule it

hubaxis:
Pending calls never started, it doesn't make sense fail it, when connection established all calls will be rescheduled. See https://github.com/grpc/grpc-dart/blob/master/lib/src/client/http2_connection.dart#L129


gustav3d:
Hows it going ?

PR #319, opened 2020-07-10
Change response from a final variable to a var

wayne-li2:
This better aligns with the gRPC docs, which heavily use this helloworld client.

The docs have you add these two lines:
`
response = await stub.sayHelloAgain(HelloRequest()..name = name);
`
`
print('Greeter client received: ${response.message}');
`

which of course will error with
`
bin/client.dart:35:5: Error: Can't assign to the final variable 'response'.
`

### DIFF
--- a/example/helloworld/bin/client.dart
+++ b/example/helloworld/bin/client.dart
@@ -30,7 +30,7 @@ Future<void> main(List<String> args) async {
   final name = args.isNotEmpty ? args[0] : 'world';
 
   try {
-    final response = await stub.sayHello(HelloRequest()..name = name);
+    var response = await stub.sayHello(HelloRequest()..name = name);
     print('Greeter client received: ${response.message}');
   } catch (e) {
     print('Caught error: $e');


mraleph:
> The docs have you add these two lines:
> `response = await stub.sayHelloAgain(HelloRequest()..name = name);`
> `print('Greeter client received: ${response.message}');`

From what I can see the docs just give the whole `main` rather than instruct to add two lines:

https://grpc.io/docs/languages/dart/quickstart/#update-the-client

But I don't think I have any strong opinion on this.

wayne-li2:
Yeah so the first line says:
```Add a call to sayHelloAgain() in bin/client.dart like this:```

client code is provided in our repo:
```
  try {
    final response = await stub.sayHello(HelloRequest()..name = name);
    print('Greeter client received: ${response.message}');
  } catch (e) {
    print('Caught error: $e');
  }
```

with the expectation to add
```
    response = await stub.sayHelloAgain(HelloRequest()..name = name);
    print('Greeter client received: ${response.message}');
```

Agreed, super minor detail, I made this PR when I was first experimenting with this repo and wasn't considering how actively it was being maintained. This is super super lower priority.

PR #322, opened 2020-07-12
Relax crypto version constraint (for Flutter)

### LINKED ISSUE #321: Can't use with Flutter because of crypto constraint
Excerpt from the [current `pubspec.yaml`](https://github.com/grpc/grpc-dart/blob/master/pubspec.yaml#L14):

```yaml
dependencies:
  crypto: ^2.1.5
```

When I try to use `grpc-dart ^2.2.0` in a simple Flutter client app (using Flutter@[stable v1.17.5](https://github.com/flutter/flutter/releases/tag/1.17.5)), I get the following dependency error:

```nocode
Running "flutter pub get" in myapp...                           
Because every version of flutter_test from sdk depends on crypto 2.1.4 and grpc 2.2.0 depends on crypto ^2.1.5, flutter_test from sdk is incompatible with grpc 2.2.0.
```

From the [crypto Changelog](https://pub.dev/packages/crypto#-changelog-tab-) we can read:

> ## 2.1.5 
> - Improve example and package description to address package site maintenance suggestions.
> ## 2.1.4 
> - BugFix: ...

Given that v2.1.5 only updated descriptions, **is there any reason not to depend on crypto ^2.1.4 instead?**

> Note: There is of course a workaround: to use a dependency override in my app, but I'd rather avoid that if possible especially since I'm putting this together for use in a simple tutorial.

chalin:
Closes #321

cc @sigurdm @mit-mit @ejona86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,14 @@
+## 2.2.0+1
+
+* Relax `crypto` version dependency constraint from `^2.1.5` to `^2.1.4`.
+
 ## 2.2.0
 
 * Added `applicationDefaultCredentialsAuthenticator` function for creating an
   authenticator using [Application Default Credentials](https://cloud.google.com/docs/authentication/production).
 * Less latency by using the `tcpNoDelay` option for sockets.
 * Support grpc-web in a non-web setting.
+
 ## 2.1.3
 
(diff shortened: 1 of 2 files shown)


chalin:
Travis is all green 😄

chalin:
Anything more to be done before this lands?

PR #328, opened 2020-07-22
Give proper api name for ChannelCredentials certificate byte list

ryanhanks:
`certificates` is a confusing name for a `List<int>` intended to hold the bytes for a single certificate, so renaming this property to `certificateBytes`.


ryanhanks:
@sigurdm I see you've been active here recently. Would you be able to review for me, or point me to someone who can?

Thanks!

Issue #329, opened 2020-07-23
gRPC error message with non ASCII UTF-8 characters are not encoded on server side.


wapa5pow:
version: 2.2.0

According to https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses , "Status-Message" should be Percent-Byte-Unencoded or Percent-Byte-Encoded.

However, grpc-dart server sends non ASCII UTF-8 (e.g. `エラー`) error message with invalid charactes.

![image](https://user-images.githubusercontent.com/5561944/88252611-d9b22b00-cce9-11ea-92d1-8d557f7c219c.png)

The correct message should be the following.

![image](https://user-images.githubusercontent.com/5561944/88252669-09613300-ccea-11ea-9846-7b7c6faa1626.png)

---

Note: with grpc-go server, the following is the result.

![image](https://user-images.githubusercontent.com/5561944/88252697-20a02080-ccea-11ea-8070-a447bba99cb5.png)

PR #330, opened 2020-07-23
Encode server error message

wapa5pow:
https://github.com/grpc/grpc-dart/issues/329

### DIFF
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -293,7 +293,8 @@ class ServerHandler_ extends ServiceCall {
     _customTrailers = null;
     outgoingTrailersMap['grpc-status'] = status.toString();
     if (message != null) {
-      outgoingTrailersMap['grpc-message'] = message;
+      outgoingTrailersMap['grpc-message'] =
+          Uri.encodeFull(message).replaceAll("%20", " ");
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/server/handler.dart
```diff
@@ -293,7 +293,13 @@ class ServerHandler_ extends ServiceCall {
     _customTrailers = null;
     outgoingTrailersMap['grpc-status'] = status.toString();
     if (message != null) {
-      outgoingTrailersMap['grpc-message'] = message;
+      try {
+        outgoingTrailersMap['grpc-message'] =
+            Uri.encodeFull(message).replaceAll("%20", " ");
+      } catch (_) {
```
sigurdm, line 299: It is not clear to me what exceptions we can get here.

https://api.dart.dev/stable/2.8.4/dart-core/Uri/encodeFull.html doesn't mention any exceptions

wapa5pow: You're right. Encoding string does not throw any exception. I fixed it.

wapa5pow:
CI says EasyCLA expected but I already signed it before ( https://github.com/grpc/grpc-dart/pull/326#issuecomment-660697252 ). Maybe, because I force pushed my branch.

sigurdm:
I'm trying to close and reopen.

PR #335, opened 2020-07-30
Remove leading slash for gRPC Web

### LINKED ISSUE #258: Use URI path in requests
Hey,
I'd like to pass a URI with a path to the client, so that I can use a proxy in front of my application on the server side to redirect to the correct service.

```dart
final channel = GrpcWebClientChannel.xhr(Uri.parse('http://localhost:6060/grpc'));
```

Now, looking at the source code after trying a couple of things, I can see that the path is simply overwritten by whatever the generated file has.
As of now I am not sure if that's a gRPC requirement or just not implemented in this project.

In line 167 I can see that the path is simply passed into the URI.
https://github.com/grpc/grpc-dart/blob/ae17e712e4f5431eb6664f0589899c77462e5475/lib/src/client/transport/xhr_transport.dart#L164-L167

Part of my pubspec.lock
```yaml
  grpc:
    dependency: "direct main"
    description:
      name: grpc
      url: "https://pub.dartlang.org"
    source: hosted
    version: "2.1.3"
```

kkweon:
so that it can respect the path prefix in the URI.

The current behavior is that the path is completely ignored because
there is a leading slash when the dart code is generated.

For example, '/protos.chat_service.ChatService/CreateStream'.

However, when it gets resolved using `Uri.resolve`,

```dart
void main() {
  final path = '/abc/def';
  final uri = Uri.parse('http://localhost:6060/grpc/');
  print(uri.resolve(path)); // http://localhost:6060/abc/def
  // /grpc/ is gone
}
```

By removing the leading slash when xhr_transport, it can work with path.
Only gRPC web supports path.

## Additional info

The following proto 

```proto
syntax = "proto3";

service Hello {
  rpc SayHello(EmptyMessage) returns (EmptyMessage);
}

message EmptyMessage {}
```

will create a path with the leading slash (/Hello/SayHello)

```dart
class HelloClient extends $grpc.Client {
  static final _$sayHello =
      $grpc.ClientMethod<$0.EmptyMessage, $0.EmptyMessage>(
          '/Hello/SayHello',
          ($0.EmptyMessage value) => value.writeToBuffer(),
          ($core.List<$core.int> value) => $0.EmptyMessage.fromBuffer(value));

  HelloClient($grpc.ClientChannel channel, {$grpc.CallOptions options})
      : super(channel, options: options);

  $grpc.ResponseFuture<$0.EmptyMessage> sayHello($0.EmptyMessage request,
      {$grpc.CallOptions options}) {
    final call = $createCall(_$sayHello, $async.Stream.fromIterable([request]),
        options: options);
    return $grpc.ResponseFuture(call);
  }
}
```

And then we initialize 

```dart
final channel =
  GrpcWebClientChannel.xhr(Uri.parse("https://k8s.kkweon.dev/grpc/"));
```

And this /grpc/ will be ignored at the current version.
The proposed fix is to change /Hello/SayHello -> Hello/SayHello for gRPC web.

And, it works for users who don't have path as well.

## Related Issues

- fixes #258

### DIFF
--- a/lib/src/client/transport/xhr_transport.dart
+++ b/lib/src/client/transport/xhr_transport.dart
@@ -134,8 +134,24 @@ class XhrClientConnection extends ClientConnection {
     request.headers['X-Grpc-Web'] = '1';
   }
 
+  // When dart codes are generated from service.proto,
+  // it has a leading slash. Therefore, it's currently not possible to use the
+  // path gets ignored. Hence, remove the leading slash.
+  //
+  // For example, we want to keep '/grpc/' path for gRPC web.
+  //
+  //   Uri.parse('http://localhost:6060/grpc/').resolve('abc/path')
+  //   // http://localhost:6060/grpc/abc/def
+  String _removeLeadingSlash(String path) {
+    if (path.startsWith("/")) {
+      return path.replaceFirst("/", "");
+    }
+    return path;
+  }
+
   @visibleForTesting
-  Request createHttpRequest(String path) => Request('POST', uri.resolve(path));
+  Request createHttpRequest(String path) =>
+      Request('POST', uri.resolve(_removeLeadingSlash(path)));
 
   @visibleForTesting
   Client createClient() => Client();
--- a/lib/src/client/transport/xhr_transport.dart
+++ b/lib/src/client/transport/xhr_transport.dart
@@ -135,8 +135,8 @@ class XhrClientConnection extends ClientConnection {
   }
 
   // When dart codes are generated from service.proto,
-  // it has a leading slash. Therefore, it's currently not possible to use the
-  // path gets ignored. Hence, remove the leading slash.
+  // it has a leading slash. Therefore, the path is ignored as it starts from
+  // the root URL. Hence, we remove the leading slash here.
   //
   // For example, we want to keep '/grpc/' path for gRPC web.
   //
--- a/test/client_tests/client_xhr_transport_test.dart
+++ b/test/client_tests/client_xhr_transport_test.dart
@@ -261,4 +261,11 @@ void main() {
       }
     }, count: expectedMessages.length));
   });
+
+  test('URL path is not ignored when creating a httpRequest', () {
+    final connection =
+        XhrClientConnection(Uri.parse("http://example.com/grpc/"));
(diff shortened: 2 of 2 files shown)


mit-mit:
This looks like a good change, cc @mraleph

mraleph:
What about `onConnectionReady` in `ClientCall`? The code there seems to build URI to pass to the metadata provider in a way that does not account for additional path components. 

I also looked at JS implementation of grpc-web and it does not seem to support this sort of functionality (though I don't understand if this is intentional or not). There is a similar issue there https://github.com/grpc/grpc-web/issues/702. 

So I am a bit hesitant accepting this - primarily because I am a) concerned that there might be other places where uri will get out of sync (like with metadata providers above b) canonical grpc-web implementation does not seem to support this. 

The benefit of implementing this is not entirely clear to me. As grpc/grpc-web#702 points out you can just use packages if you want to have a fixed prefix to use in your server configuration.

kkweon:
the package seems like a hack to me..

Although I agree with the concern.
I'd like to point out it works in JS(TypeScript) gRPC web implementations because they don't use path resolve.

```typescript
  sayHello(
    request: EmptyMessage,
    metadata: grpcWeb.Metadata | null,
    callback: (err: grpcWeb.Error,
               response: EmptyMessage) => void) {
    return this.client_.rpcCall(
      this.hostname_ +
        '/Hello/SayHello',
      request,
      metadata || {},
      this.methodInfoSayHello,
      callback);
  }
```

They simply append.

mit-mit:
cc @jtattermusch can someone from the grpc core team advice is what the right behavoir is here?

jtattermusch:
> cc @jtattermusch can someone from the grpc core team advice is what the right behavoir is here?

@stanleycheung please advise.

mit-mit:
Kind reminder @stanley-cheung

mraleph:
Closing as assumed stale for now.

RyanSusana:
Any update on this?

hermantai:
Hi, Vyacheslav, is this closed by mistake? The issue still exists, so it's not stale. Thanks.

mraleph:
It got closed because we did not get any follow up answer from the core grpc team and I personally have no opinion on whether this should be supported or not.

Issue #286, opened 2020-03-23
Documentation Error for Dart GRPC Basics


stephengrice:
Hello, I got confused when following the GRPC Basics article about writing a client [here](https://grpc.io/docs/tutorials/basic/dart/#creating-a-stub). I had to examine the [example source code](https://github.com/grpc/grpc-dart/blob/master/example/route_guide/lib/src/client.dart) to figure out what the problem was. The variable referenced as "client" in the tutorial is later changed to "stub" without any explanation. I think this is an easy fix, just make them both match. Here is what I mean:

Code snippet 1 - Variable referred to as **client**:
```
final client = RouteGuideClient(channel,
    options: CallOptions(timeout: Duration(seconds: 30)));
```
Code snippet 2 - Variable referred to as **stub**:
```
final point = Point()
  ..latitude = 409146138
  ..longitude = -746188906;
final feature = await stub.getFeature(point));
```

<version of the grpc-dart packages used; see your `pubspec.lock` file>
n/a

## Repro steps

1. Visit the [article](https://grpc.io/docs/tutorials/basic/dart/#creating-a-stub).

2. View the code snippets side-by-side.

Expected result: **client** and **stub** should have been the same name

Actual result: I had no idea what **stub** was referring to until digging deeper.

## Details

Referred to [example source](https://github.com/grpc/grpc-dart/blob/master/example/route_guide/lib/src/client.dart) to find the answer. To match this, both references should be changed to **stub**.

JarrodCColburn:
@stephengrice 
Hey, this might be a good suggestion it's been a while since I've done the tutorial. 
But the documentation is actually in the repo grpc.io (not here on grpc-dart) and the file is here https://github.com/grpc/grpc.io/blob/master/content/docs/languages/dart/basics.md

Could you please edit and open a PR there and close this one. thanks

stephengrice:
Thank you Jarrod. I opened an issue on the repository as you suggested. Closing this one out.

PR #350, opened 2020-09-15
Example: route guide client cleanup

chalin:
The `channel` variable should be a local variable rather than a member of `Client`. (This is in part a followup to #286 and https://github.com/grpc/grpc.io/issues/386.)

@sigurdm @mit-mit 

I've manually tested the route guide example to ensure that it still works as intended.

Related: https://github.com/grpc/grpc.io/pull/395

### DIFF
--- a/example/route_guide/lib/src/client.dart
+++ b/example/route_guide/lib/src/client.dart
@@ -22,11 +22,10 @@ import 'generated/route_guide.pb.dart';
 import 'generated/route_guide.pbgrpc.dart';
 
 class Client {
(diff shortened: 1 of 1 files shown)


mit-mit:
lgtm, @mraleph can you give this an eng review?

mraleph:
LGTM. There is not much to review really. (I don't have merge/review rights yet though so you will have to merge @mit-mit )

chalin:
Thank you both for the quick review and merge!

PR #352, opened 2020-09-17
Regenerate certificates used by tests.

mraleph:
Current certificates don't pass verification on Mac OS X Catalina
making it impossible to run test suite on Mac laptops.

PR #353, opened 2020-09-18
Remove check for SecurityContext.alpnSupported in security#createSecurityContext

wayne-li2:
Looks like SecurityContext.alpnSupported and always returns true in Dart SDK >=1.25.0. The pubspec currently requires >=2.2.0 so this should be fine to remove.

I pulled in commits from #352 to run unit tests, and all looked good.


mraleph:
Thanks!

PR #354, opened 2020-09-18
Add a simple end-to-end gRPC-web test

mraleph:
We did not have any before which allowed for regressions like #306 to
slip through. Unfortunately we can't test gRPC-web implementation
in pure Dart because we don't have a server side implementation of
the protocol. Instead we add a dependency on the third party
gRPC-web proxy (grpcwebproxy by Improbable Engineering - the choice made
purely based on the simplicity of installation) which forwards all
request gRPC server (written in Dart).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
 before_install:
   - if [ $TRAVIS_OS_NAME = windows ]; then choco install googlechrome ; fi
   - ./tool/install-grpcwebproxy.sh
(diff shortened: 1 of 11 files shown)


joe-getcouragenow:
@mraleph nice.

Improbable is also what many use to avoid the envoy dependency.
i fing the dart grpc working with both btw

mraleph:
@nichite FYI, I am going to TBR land this because this PR is tests only and I would like to have additional testing before I am going to land anything else.

PR #360, opened 2020-09-25
Beef up exception handling in gRPC code.

nichite:
This PR syncs over some improvements to exception plumbing we did internally.


mraleph (approved):
If possible consider adding a test which verifies that stackTrace is populated with something meaningful

PR #363, opened 2020-09-29
Add client validation to server credentials

nichite:
Sync from internal that gives some more client validation ability to server credentials.

Also, expose `MetadataProvider` in the service API to provide utility functions in various Flutter libraries.

### DIFF
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -120,12 +120,8 @@ class Server {
     final securityContext = security?.securityContext;
     if (securityContext != null) {
       _secureServer = await SecureServerSocket.bind(
-          address ?? InternetAddress.anyIPv4,
-          port ?? 443,
-          securityContext,
-          backlog: backlog,
-          shared: shared,
-          v6Only: v6Only);
+          address ?? InternetAddress.anyIPv4, port ?? 443, securityContext,
+          backlog: backlog, shared: shared, v6Only: v6Only);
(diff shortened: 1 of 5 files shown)


lib/src/server/server.dart
```diff
@@ -135,10 +160,14 @@ class Server {
         _connections.remove(connection);
       });
     }, onError: (error) {
-      print('Socket error: $error');
+      _printSocketError(error);
```
mraleph, line 163: Maybe be this can just be `onError: _printSocketError`?

lib/src/server/server.dart
```diff
@@ -135,10 +160,14 @@ class Server {
...
     });
   }
 
+  void _printSocketError(String error) {
```
mraleph, line 167: This should be `Object error` (because onError handler would pass non-strings in here). This actually caused some breakage internally when first committed, the fact that it is not causing issue here means our test coverage is not stellar

mraleph (approved):
LGTM with comments

lib/src/server/server.dart
nichite: That wasn't working before with the incorrect typing. But now that `_printSocketError` accepts an `Object`, it works. Updated.

lib/src/server/server.dart
nichite: Done.

Issue #339, opened 2020-08-18
Client is stuck in infinite attempt to connect due to SocketException: OS Error: No route to host


Nimrodda:
I'm working on error handling of my app (Flutter) and I noticed that if I shutdown the server completely, the client is stuck in a connection attempt loop with an error `SocketException: OS Error: No route to host, errno = 113, address = 10.0.0.18, port = 43926`. I've searched the web and investigated the source code and didn't see a way to catch this error. The endpoint is a server side streaming.

Version 2.2.0

## Repro steps

1. Put random IP and port when constructing the channel
2. Try to connect to a terminated server

Expected result: 
A way to catch the error.

Actual result:
No way to catch the error (at least after taking a look into the code)

## Details

While debugging, I see that this code throws an exception:
```
Future<ClientTransportConnection> connectTransport() async {
    final securityContext = credentials.securityContext;
    Socket socket = await Socket.connect(host, port); <--- HERE
```

Which is then caught here:
```
connectTransport().then((transport) async {
      _currentReconnectDelay = null;
      _transportConnection = transport;
      _connectionLifeTimer
        ..reset()
        ..start();
      transport.onActiveStateChanged = _handleActiveStateChanged;
      _setState(ConnectionState.ready);

      if (_hasPendingCalls()) {
        // Take all pending calls out, and reschedule.
        final pendingCalls = _pendingCalls.toList();
        _pendingCalls.clear();
        pendingCalls.forEach(dispatchCall);
      }
    }).catchError(_handleConnectionFailure); <---- HERE
```

but the problem is that `_handleConnectionFailure` is stuck in a loop with no way for me - the user of the library - to catch the error and stop trying to connect.

```
void _handleConnectionFailure(error) {
    _transportConnection = null;
    if (_state == ConnectionState.shutdown || _state == ConnectionState.idle) {
      return;
    }
    // TODO(jakobr): Log error.
    _cancelTimer();
    _setState(ConnectionState.idle);
    _connect(); <-- Stuck in a loop
  }
```

PR #364, opened 2020-10-01
Create gRPC servers and clients with TransportConnections.

nichite:
This allows callers to propvide their own transport configuration, such
as their own implementation of streams and sinks instead of sockets.

* Extract ConnectionServer from Server to support ServerTransportConnection
* Add ClientTransportConfig to support ClientTransportConnection
* Plumb ClientTransportConfig from Channel to Http2ClientConnection

### DIFF
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -80,22 +80,14 @@ class ServerTlsCredentials extends ServerCredentials {
 /// A gRPC server.
 ///
 /// Listens for incoming RPCs, dispatching them to the right [Service] handler.
-class Server {
-  final Map<String, Service> _services = {};
-  final List<Interceptor> _interceptors;
-
+class Server extends ConnectionServer {
   ServerSocket _insecureServer;
   SecureServerSocket _secureServer;
-  final _connections = <ServerTransportConnection>[];
 
   /// Create a server for the given [services].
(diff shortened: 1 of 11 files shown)


nichite:
Looks like some of the tests are timing out--I'm not sure why this isn't happening internally. Will need to look into it.

mraleph:
This is due to this commit https://github.com/grpc/grpc-dart/commit/9ed03b6b969b9171d5b218e0ac308ae127c56984 which causes transport to endlessly attempt reconnecting on connection failures. This also causes https://github.com/grpc/grpc-dart/issues/339 - maybe we should just revert the commit that introduced reconnection logic for now.

lib/src/server/server.dart
```diff
@@ -190,3 +160,49 @@ class Server {
     _secureServer = null;
   }
 }
+
+/// A gRPC server that serves via provided [ServerTransportConnection]s.
```
mraleph, line 164: I  find the ordering a bit confusing - I would expect base class to come ahead of child in the file

mraleph (approved):

mhoad:
@nichite can I ask a dumb newbie question? Does this commit open the door to a non XHR based grpc-web future by opening up new possibilities like GRPC over Websockets and Streams or am I totally misreading this?

PR #366, opened 2020-10-02
Revert "Not to fail pending requests, just reschedule it (#303)"

mraleph:
This reverts commit 9ed03b6b969b9171d5b218e0ac308ae127c56984.

This change prevents client from gracefully handling connection issues and instead just makes it infinitely trying to reconnect (#339).


nichite (approved):
Thanks!

PR #369, opened 2020-10-13
Create SecureSocket directly instead of upgrading.

nichite:
The current behavior creates an insecure socket and then attempts to upgrade it. Since socket-level checks have been rolled back in the Dart SDK, no platform does this level of check--it proves to be very tricky to handle the case of upgraded insecure sockets.

Instead, we expose the network policy via `isInsecureConnectionAllowed` function which is used by dart:io HttpClient. Any client that deals with sockets directly should also use this function to disallow insecure connections where applicable.

This change makes sure gRPC follows suit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 ## 2.6.1
 
-* Create Securesocket directly when credentials.authority is not present
-  instead of always creating an insecure socket first and upgrading.
+* Create `Securesocket` directly when `ChannelCredentials.authority` is not
+  present instead of always creating an insecure socket first and upgrading.
 * Throw on insecure connections when disallowed by clients dealing with
-  socket code.
+  socket code, if set via `isInsecureConnectionAllowed`.
 
 ## 2.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 2.6.1
+
+* Create Securesocket directly when credentials.authority is not present
+  instead of always creating an insecure socket first and upgrading.
+* Throw on insecure connections when disallowed by clients dealing with
+  socket code.
+
 ## 2.6.0
 
 * Create gRPC servers and clients with [Server|Client]TransportConnnection.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 2.6.1
+
+* Create Securesocket directly when credentials.authority is not present
```
mraleph, line 3: `SecureSocket`, `ChannelCredentials.authority`

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+  instead of always creating an insecure socket first and upgrading.
+* Throw on insecure connections when disallowed by clients dealing with
```
mraleph, line 5: This should probably at least refer to `isInsecureConnectionAllowed`.

lib/src/client/http2_connection.dart
```diff
@@ -308,21 +308,36 @@ class _SocketTransportConnector implements ClientTransportConnector {
...
     final securityContext = _options.credentials.securityContext;
-    _socket = await Socket.connect(_host, _port);
-    // Don't wait for io buffers to fill up before sending requests.
-    _socket.setOption(SocketOption.tcpNoDelay, true);
-    if (securityContext != null) {
-      // Todo(sigurdm): We want to pass supportedProtocols: ['h2'].
-      // http://dartbug.com/37950
-      _socket = await SecureSocket.secure(_socket,
+    if (securityContext == null) {
+      if (!isInsecureConnectionAllowed(_host)) {
```
mraleph, line 312: If possible please add at least a simple test.

mraleph (approved):

CHANGELOG.md
nichite: Done.

CHANGELOG.md
nichite: Done.

lib/src/server/server.dart
```diff
@@ -174,6 +174,11 @@ class Server extends ConnectionServer {
       socket.setOption(SocketOption.tcpNoDelay, true);
       final connection = ServerTransportConnection.viaSocket(socket,
           settings: http2ServerSettings);
+      if (security != null && !security.validateClient(socket)) {
```
mraleph, line 177: I guess we should write a simple test for this as well given how easy we have accidentally lost it (I did not notice it in my review, sorry).

PR #371, opened 2020-10-16
fix: duplicate headers in grpc-web

zbigniew-mandziejewicz-3r:
`content-type`, `x-user-agent` and `x-grpc-web` headers are sent twice, causing issues with Ambassador.
Removed header setting from `_initRequest` since metadata is already populated in `makeRequest`.


mraleph:
This is already done in https://github.com/grpc/grpc-dart/pull/369 which coincidentally I think should be the last internal-to-external PR which lands before we start taking external PRs (waiting on @nichite to confirm).

zbigniew-mandziejewicz-3r:
Thanks for heads-up @mraleph

PR #372, opened 2020-10-16
Ensure CallOptions merge preserves Web specific options

zbigniew-mandziejewicz-3r:
When WebCallOptions are merged with CallOptions data is lost, this PR preserves the original WebCallOptions flags and returns correct type.


shaxbee:
@mraleph We should land this to make interceptors that modify `callOptions` work with grpc-web clients.

mraleph:
Please add a test to make sure we don't break it again. I would be happy to merge afterwards. 

(We are looking at ways to mitigate Travis slowness in the meantime)

shaxbee:
> Please add a test to make sure we don't break it again. I would be happy to merge afterwards.

Done.

shaxbee:
@mraleph Please trigger rebuild if chrome issue are resolved.

PR #383, opened 2020-11-03
Create dart.yml

mit-mit:


mraleph:
For some reason workflow is not rerunning on the pushed changes, so it is hard to say whether it is working or not.

mit-mit:
Hmm OK. Can we try to merge this, and then I'll send a PR with fixes as needed?

mit-mit:
We seem to be in some odd intermittent state where even new PRs (e.g. https://github.com/grpc/grpc-dart/pull/385) don't trigger Actions. I speculate that we have to merge this base `.yml` and then send PRs to change it.

PR #387, opened 2020-11-03
Update dart.yml

mraleph:
grpc master branch name is `master` not `main`.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -25,9 +25,7 @@ jobs:
       - name: Check formatting
         run: dart format --output=none --set-exit-if-changed .
       - name: Analyze code (lib and test)
-        run: |
-          dart analyze lib
-          dart analyze test
+        run: dart analyze .
       - name: Analyze code (introp and examples)
         run: |
           for example in interop example/*/; do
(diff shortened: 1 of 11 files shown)


mit-mit:
Looks like the failures are in the `example` dir. The Travis CI script has this comment:

```
include:
  # Wish we could exclude `example` in `analysis_options.yaml` but it seems
  # blocked by https://github.com/dart-lang/sdk/issues/26212
  - dart: dev
    dart_task: dartfmt
    script:
    - pub get
    - dartanalyzer lib test
```

That bug is fixed now. Can you try to add this to `analysis_options.yaml`?

```
# Skip analysis of the example dir
analyzer:
  exclude:
  - example/**
```

mraleph:
@mit-mit I am just going to try to replicate what Travis did - analyze each example independently

mraleph:
@mit-mit ha, `dart analyze ` has different CLI interface from `dartanalyzer` - why is that?

`dartanalyzer lib test` works, but `dart analyze lib test` does not.

mit-mit:
For the core code (everything but /example), I think it'll work if you change:
```
      - name: Analyze code (lib and test)
        run: dart analyze lib test
```

to:
```
      - name: Analyze code (lib and test)
        run: dart analyze .
```

PR #390, opened 2020-11-03
Enable Web tests and make sure to install right channel

mraleph:


tool/install-grpcwebproxy.sh
```diff
@@ -16,9 +17,23 @@ case $TRAVIS_OS_NAME in
...
+    VARIANT=win64.exe
+    SUFFIX=.exe
+    WGET=C:/msys64/usr/bin/wget.exe
+    ;;
+esac
+
 BINARY=grpcwebproxy-${VERSION}-${VARIANT}
 
-wget https://github.com/improbable-eng/grpc-web/releases/download/${VERSION}/${BINARY}.zip -O /tmp/grpcwebproxy.zip
+${WGET} https://github.com/improbable-eng/grpc-web/releases/download/${VERSION}/${BINARY}.zip -O /tmp/grpcwebproxy.zip
```
mit-mit, line 36: @mraleph in the install-dart I'm working on, I'm using curl and that seems to work, e.g.:
`curl --connect-timeout 15 --retry 5 "$URL" > "${HOME}/dartsdk.zip"`

mraleph: I am going to just exclude Windows and Mac from browser testing - I don't think this is needed to begin with. (Maybe we could include Mac Safari for coverage or something like that)

PR #398, opened 2020-11-12
Cleanups

mraleph:


mraleph:
TBR @mkustermann (this is just a mechanical change)

Issue #206, opened 2019-06-26
grpc-web: Gracefully close a streaming connection


ligustah:
Calling `cancel()` on the `ResponseStream` returned by a streaming RPC throws an exception.

version: 2.0.0+1

## Repro steps
Don't have separate code for this right now, but could prepare something if required.

1. Call a server-side streaming RPC
2. Receive messages
3. Call `cancel()` on the `ResponseStream`

Expected result: The connection is closed is closed and I can handle any possible exceptions.

Actual result: The code will throw two exceptions, only one of which I can catch.

## Details

I can see why adding an error to the stream when calling `cancel()` might be useful, but somewhere in the code it tries to add another error to one of the internal streams after that one was closed, which triggers the bad state exception seen below.

```
EXCEPTION: Bad state: Cannot add event after closing
STACKTRACE: 
dart:sdk_internal                                                                                addError
package:grpc/src/client/packages/grpc/src/client/call.dart 305:16                                [_onRequestError]
package:grpc/packages/grpc/src/client/transport/xhr_transport.dart 61:13                         <fn>

...

EXCEPTION: gRPC Error (1, Cancelled by client.)
STACKTRACE: 
package:grpc/src/client/packages/grpc/src/client/call.dart 324:18                                cancel
package:grpc/src/client/packages/grpc/src/client/common.dart 85:34                               cancel
```

Potentially related: #111

softkot:
I am confirming this annoying bug in 2.3.0

mraleph:
Duplicate of #67

aragonet:
The same is happening here on web with grpc version 2.8.0

PR #414, opened 2020-12-17
Prevent an exception to be thrown when a web connection stream is closed

Herrera93:
- Related #206

### DIFF
--- a/lib/src/client/transport/xhr_transport.dart
+++ b/lib/src/client/transport/xhr_transport.dart
@@ -44,6 +44,7 @@ class XhrTransportStream implements GrpcTransportStream {
   final StreamController<ByteBuffer> _incomingProcessor = StreamController();
   final StreamController<GrpcMessage> _incomingMessages = StreamController();
   final StreamController<List<int>> _outgoingMessages = StreamController();
+  bool _terminated = false;
 
   @override
   Stream<GrpcMessage> get incomingMessages => _incomingMessages.stream;
@@ -59,7 +60,7 @@ class XhrTransportStream implements GrpcTransportStream {
         .listen((data) => _request.send(data), cancelOnError: true);
 
     _request.onReadyStateChange.listen((data) {
-      if (_incomingMessages.isClosed) {
+      if (_incomingMessages.isClosed || _terminated) {
         return;
       }
       switch (_request.readyState) {
@@ -150,6 +151,7 @@ class XhrTransportStream implements GrpcTransportStream {
 
   @override
   Future<void> terminate() async {
+    _terminated = true;
     _close();
     _request.abort();
   }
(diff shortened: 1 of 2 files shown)


mraleph:
Hi @Herrera93,

Thank you for the contribution. 

Please rebase your branch - that should unblock testing.  

We would also need tests for this specific test to prevent regressions in the future.

mraleph:
Thank you for adding a test.

I have looked at things more closely and I can see that the root cause of the issue is that we are checking whether `_incommingMessages` are closed, but we should be rather checking that `_incommingProcessor` is closed (when processor is closed it propagates closure to messages stream - but this happens asynchronously so it might race with other events, which causes this bug).

I think the proper fix is like this (no need to introduce `_terminated`):

```diff
diff --git a/lib/src/client/transport/xhr_transport.dart b/lib/src/client/transport/xhr_transport.dart
index 89259c5..9244491 100644
--- a/lib/src/client/transport/xhr_transport.dart
+++ b/lib/src/client/transport/xhr_transport.dart
@@ -44,7 +44,6 @@ class XhrTransportStream implements GrpcTransportStream {
   final StreamController<ByteBuffer> _incomingProcessor = StreamController();
   final StreamController<GrpcMessage> _incomingMessages = StreamController();
   final StreamController<List<int>> _outgoingMessages = StreamController();
-  bool _terminated = false;
 
   @override
   Stream<GrpcMessage> get incomingMessages => _incomingMessages.stream;
@@ -60,7 +59,7 @@ class XhrTransportStream implements GrpcTransportStream {
         .listen((data) => _request.send(data), cancelOnError: true);
 
     _request.onReadyStateChange.listen((data) {
-      if (_incomingMessages.isClosed || _terminated) {
+      if (_incomingProcessor.isClosed) {
         return;
       }
       switch (_request.readyState) {
@@ -81,7 +80,7 @@ class XhrTransportStream implements GrpcTransportStream {
     });
 
     _request.onError.listen((ProgressEvent event) {
-      if (_incomingMessages.isClosed) {
+      if (_incomingProcessor.isClosed) {
         return;
       }
       _onError(GrpcError.unavailable('XhrConnection connection-error'),
@@ -90,7 +89,7 @@ class XhrTransportStream implements GrpcTransportStream {
     });
 
     _request.onProgress.listen((_) {
-      if (_incomingMessages.isClosed) {
+      if (_incomingProcessor.isClosed) {
         return;
       }
       // Use response over responseText as most browsers don't support
@@ -151,7 +150,6 @@ class XhrTransportStream implements GrpcTransportStream {
 
   @override
   Future<void> terminate() async {
-    _terminated = true;
     _close();
     _request.abort();
   }
```

Could you apply this patch on top of your branch (normally I would be able to just commit a patch myself into the PR - but you created PR from your master branch which prevents me from doing this, so I have to ask you do it).

After that I can land the PR.

long1eu:
can we have this published?

mraleph:
Published 2.9.0

PR #419, opened 2021-01-08
Allow to make future chain request with ResponseFuture

### LINKED ISSUE #413: Interceptor for handling response
Can we use ClientInterceptor for handling common use cases or errors from the response?

rockerhieu:
Closes #413

This PR allows to make future chain request with `ResponseFuture`:

```dart
  @override
  ResponseFuture<R> interceptUnary<Q, R>(ClientMethod<Q, R> method, Q request,
      CallOptions options, ClientUnaryInvoker<Q, R> invoker) {
    _invocations.add(InterceptorInvocation(_id, ++_unary, _streaming));
    return ResponseFuture.wrap(
            Future.delayed(Duration(seconds: 1), () => 'dummy'))
        .then((_) => invoker(method, request, _inject(options)))
        .then((foo) async => foo)
        .whenComplete(() => 'complete')
        .then((bar) => bar)
        .catchError((e, s) => print('$e at $s'))
        .timeout(Duration(seconds: 5));
  }
```

### DIFF
--- a/lib/src/client/common.dart
+++ b/lib/src/client/common.dart
@@ -44,8 +44,7 @@ abstract class Response {
 /// A gRPC response producing a single value.
 class ResponseFuture<R> extends DelegatingFuture<R>
     with _ResponseMixin<dynamic, R> {
-  @override
-  final ClientCall<dynamic, R> _call;
+  final ClientCall _call;
 
   static R _ensureOnlyOneResponse<R>(R? previous, R element) {
     if (previous != null) {
@@ -63,6 +62,45 @@ class ResponseFuture<R> extends DelegatingFuture<R>
       : super(_call.response
             .fold<R?>(null, _ensureOnlyOneResponse)
             .then(_ensureOneResponse));
+
+  ResponseFuture._wrap(Future<R> future, {ClientCall clientCall})
+      : _call = clientCall,
+        super(future);
+
+  /// `clientCall` maybe be lost when converting from Future to ResponseFuture
+  static ResponseFuture<T> wrap<T>(Future<T> future, {ClientCall clientCall}) {
+    return ResponseFuture._wrap(
+      future,
+      clientCall: (_unwrap(future) ?? clientCall),
+    );
+  }
+
+  static ClientCall _unwrap(Future future) =>
+      future is ResponseFuture ? future._call : null;
+
+  @override
+  ResponseFuture<S> then<S>(FutureOr<S> Function(R p1) onValue,
+      {Function onError}) {
+    return wrap(super.then(onValue, onError: onError), clientCall: _call);
+  }
+
+  @override
+  ResponseFuture<R> catchError(Function onError,
+      {bool Function(Object error) test}) {
+    return wrap(super.catchError(onError, test: test), clientCall: _call);
+  }
+
+  @override
+  ResponseFuture<R> whenComplete(FutureOr Function() action) {
+    return wrap(super.whenComplete(action), clientCall: _call);
+  }
+
+  @override
+  ResponseFuture<R> timeout(Duration timeLimit,
+      {FutureOr<R> Function() onTimeout}) {
+    return wrap(super.timeout(timeLimit, onTimeout: onTimeout),
+        clientCall: _call);
+  }
 }
 
 /// A gRPC response producing a stream of values.
@@ -78,7 +116,7 @@ class ResponseStream<R> extends DelegatingStream<R>
 }
 
 abstract class _ResponseMixin<Q, R> implements Response {
-  ClientCall<Q, R> get _call;
+  ClientCall get _call;
 
   @override
   Future<Map<String, String>> get headers => _call.headers;
@@ -87,5 +125,5 @@ abstract class _ResponseMixin<Q, R> implements Response {
   Future<Map<String, String>> get trailers => _call.trailers;
 
   @override
-  Future<void> cancel() => _call.cancel();
+  Future<void> cancel() => _call?.cancel();
 }
--- a/test/client_tests/client_interceptor_test.dart
+++ b/test/client_tests/client_interceptor_test.dart
@@ -1,7 +1,7 @@
 import 'package:grpc/grpc.dart';
 import 'package:grpc/src/client/interceptor.dart';
-import 'package:test/test.dart';
 import 'package:http2/transport.dart';
+import 'package:test/test.dart';
 
 import '../src/client_utils.dart';
 import '../src/utils.dart';
@@ -32,8 +32,14 @@ class FakeInterceptor implements ClientInterceptor {
   ResponseFuture<R> interceptUnary<Q, R>(ClientMethod<Q, R> method, Q request,
       CallOptions options, ClientUnaryInvoker<Q, R> invoker) {
     _invocations.add(InterceptorInvocation(_id, ++_unary, _streaming));
-
-    return invoker(method, request, _inject(options));
+    return ResponseFuture.wrap(
+            Future.delayed(Duration(seconds: 1), () => 'dummy'))
+        .then((_) => invoker(method, request, _inject(options)))
+        .then((foo) async => foo)
+        .whenComplete(() => 'complete')
+        .then((bar) => bar)
+        .catchError((e, s) => print('$e at $s'))
+        .timeout(Duration(seconds: 5));
   }
 
   @override


rockerhieu:
@mraleph hi, do you have a chance to take a look at this PR?

lib/src/client/common.dart
```diff
@@ -58,10 +58,51 @@ class ResponseFuture<R> extends DelegatingFuture<R>
...
+
+  ResponseFuture._clone(Future<R> future, {ClientCall clientCall})
+      : _call = clientCall,
+        super(future);
+
+  ResponseFuture<S> copy<S>(Future<S> future) =>
+      ResponseFuture._clone(future, clientCall: _getClientCall(future, _call));
+
+  /// clientCall maybe be lost when converting from Future to ResponseFuture
+  static ResponseFuture<T> wrap<T>(Future<T> future, {ClientCall clientCall}) {
```
mraleph, line 75: Do we actually have any use for `wrap` beyond testing?

lib/src/client/common.dart
```diff
@@ -58,10 +58,51 @@ class ResponseFuture<R> extends DelegatingFuture<R>
...
+    return ResponseFuture._clone(
+      future,
+      clientCall: (_getClientCall(future, clientCall)),
+    );
+  }
+
+  static ClientCall _getClientCall(Future future, [ClientCall fallback]) {
```
mraleph, line 82: I think it would be easier to read if you do 

```dart
static ClientCall<dynamic, R> _unwrap<R>(Future<R> future) =>
   future is ResponseFuture ? future._call : null;
```

and then make call-sites `_unwrap(future) ?? clientCall`.

lib/src/client/common.dart
```diff
@@ -75,7 +116,7 @@ class ResponseStream<R> extends DelegatingStream<R>
 }
 
 abstract class _ResponseMixin<Q, R> implements Response {
-  ClientCall<Q, R> get _call;
+  ClientCall get _call;
```
mraleph, line 119: We should not loose type safety.

lib/src/client/common.dart
```diff
@@ -84,5 +125,5 @@ abstract class _ResponseMixin<Q, R> implements Response {
   Future<Map<String, String>> get trailers => _call.trailers;
 
   @override
-  Future<void> cancel() => _call.cancel();
+  Future<void> cancel() => _call?.cancel();
```
mraleph, line 128: I think `_call` becoming nullable is very unfortunate. Can we avoid that? I think it will make NNBD migration harder.

mraleph (changes requested):
Some initial review comments. 

Thank you for the contribution!

lib/src/client/common.dart
rockerhieu: We need a way to convert `Future` to `ResponseFuture` and vice versa. You can see we use it in the very first step in this example:

```dart
  @override
  ResponseFuture<R> interceptUnary<Q, R>(ClientMethod<Q, R> method, Q request,
      CallOptions options, ClientUnaryInvoker<Q, R> invoker) {
    _invocations.add(InterceptorInvocation(_id, ++_unary, _streaming));
    return ResponseFuture.wrap(
            Future.delayed(Duration(seconds: 1), () => 'dummy'))
        .then((_) => invoker(method, request, _inject(options)))
        .then((foo) async => foo)
        .whenComplete(() => 'complete')
        .then((bar) => bar)
        .catchError((e, s) => print('$e at $s'))
        .timeout(Duration(seconds: 5));
  }```

File: lib/src/client/common.dart [hunk 5]
```diff
@@ -58,10 +58,51 @@ class ResponseFuture<R> extends DelegatingFuture<R>
...
 
-  ResponseFuture(this._call)
-      : super(_call.response
+  ResponseFuture(ClientCall<dynamic, R> call)
+      : _call = call,
+        super(call.response
             .fold(null, _ensureOnlyOneResponse)
             .then(_ensureOneResponse));
+
+  ResponseFuture._clone(Future<R> future, {ClientCall clientCall})
```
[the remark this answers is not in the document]
rockerhieu: Sure, I can change that

lib/src/client/common.dart
rockerhieu: `.then(...)` allows to transform a `ResponseFuture<R>` to a `ResponseFuture<S>` which will lose type safety. Do you have any suggestion?

```
  @override
  ResponseFuture<S> then<S>(FutureOr<S> Function(R p1) onValue, {Function onError});
```

lib/src/client/common.dart
rockerhieu: As `Future` doesn't have `_call`, converting a `Future` to a `ResponseFuture` will make `_call == null`. Let me revisit the change once more time to find a better approach.

lib/src/client/common.dart
```diff
@@ -44,7 +44,7 @@ abstract class Response {
 /// A gRPC response producing a single value.
 class ResponseFuture<R> extends DelegatingFuture<R>
     with _ResponseMixin<dynamic, R> {
-  final ClientCall<dynamic, R> _call;
+  final ClientCall _call;
```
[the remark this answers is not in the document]
rockerhieu: Because of this: https://github.com/grpc/grpc-dart/pull/419#discussion_r563358990

lib/src/client/common.dart
rockerhieu: SGTM, let me update

lib/src/client/common.dart
rockerhieu: Another thing we can do is having a convention that the final `ResponseFuture` in the chain must have a valid `_call`.

rockerhieu:
Btw, as `interceptUnary` returns a `ResponseFuture` there is no way to mark `interceptUnary` as `async` as `await` always returns a `Future`.

rockerhieu:
@mraleph I know that this PR is quite outdated with recent changes about NNBD support. But what do you think about making `_call` optional?

mraleph:
I think original gRPC library design is really gnarly here, it's kinda painting us into the corner. 

For example, now that I look at it with fresh-eyes I realise that in your example: 

```dart
  @override
  ResponseFuture<R> interceptUnary<Q, R>(ClientMethod<Q, R> method, Q request,
      CallOptions options, ClientUnaryInvoker<Q, R> invoker) {
    _invocations.add(InterceptorInvocation(_id, ++_unary, _streaming));
    return ResponseFuture.wrap(
            Future.delayed(Duration(seconds: 1), () => 'dummy'))
        .then((_) => invoker(method, request, _inject(options)))
        .then((foo) async => foo)
        .whenComplete(() => 'complete')
        .then((bar) => bar)
        .catchError((e, s) => print('$e at $s'))
        .timeout(Duration(seconds: 5));
  }
```

interceptor simply returns a `ResponseFuture` with a `null` call and the whole chaining exercise is kinda for nothing, nobody is going to see that call anyway (and other `Response` APIs are going to be "broken" by such interceptor).

So I guess the question here is: do we want to find a way to support asynchronous actions before calling `invoker`? I think that is the hard part, so my answer would be no. 

In that case `call` is guaranteed to never be null, and we can remove public `wrap`. Basically the only way to get `ReponseFuture` is to call `invoker` or to call a method on `ResponseFuture`.

lib/src/client/common.dart
mraleph: I realized that we don't actually use the type of `call` for anything. So I suggest the following change: it is indeed fine to drop types here, but the we should add types in default constructor to keep initial type safety: 

```
  ResponseFuture(ClientCall<dynamic, R> call) : _call = call, super(call.response
            .fold<R?>(null, _ensureOnlyOneResponse)
            .then(_ensureOneResponse));

```

amondnet:
@rockerhieu  @mraleph Hi, 

`future.then()` can returns `R` instead of `T`. 
```dart
Future<R> then<R>(FutureOr<R> onValue(T value), {Function? onError});
```
Should `ResponseFuture` 's `then` also return `R`? 
```dart
final response = stub.sayHello(request).then( (HelloReply reply) => 'success');
```

Wouldn't it be better to create a new method ? 

```dart
ResponseFuture<T> responseThen<T>(FutureOr<T> onValue(T value), {Function? onError});
```

```dart
stub.sayHello(request)
    .responseThen( (HelloReply reply) => reply);


```

mraleph:
> Is it possible to change the type of `ClientCall`? Or should I block type conversion?

The type of the call does not really matter. See my comment above: https://github.com/grpc/grpc-dart/pull/419#discussion_r645446584

I think the biggest question for this PR is how to correctly support asynchronous operations before `invoker` is called and make sure that returned future gets access to `ClientCall` when that is available.

kengu:
This PR enables features that are not practical to do another way, like transforming general grpc errors to typed exceptions. One workaround could be to use stream transformations, but it seems a bit excessive.

esenmx:
The response type shouldn't be changed over interceptor, it's obviously very bad idea, on top of that we lose type-safety.

Interceptors should behave as complementary like `refreshing tokens`, then `retrying the failed request`, which requires an `async` function body.

rockerhieu:
> The response type shouldn't be changed over interceptor, it's obviously very bad idea, on top of that we lose type-safety.
> 
> Interceptors should behave as complementary like `refreshing tokens`, then `retrying the failed request`, which requires an `async` function body.
> 
> Please just let `wrap` extension method with `type-safety` that's exactly what all we need.
> If you need to change response type over interceptor, please consider using a `sum type` with `oneof` keyword.

@esenmx can you elaborate on the `wrap` extension approach? Let's take one simple use case like measuring the time it takes for `invoker(method, request, _inject(options))` to finish in `interceptUnary`. No response type changed.

esenmx:
> > The response type shouldn't be changed over interceptor, it's obviously very bad idea, on top of that we lose type-safety.
> > Interceptors should behave as complementary like `refreshing tokens`, then `retrying the failed request`, which requires an `async` function body.
> > Please just let `wrap` extension method with `type-safety` that's exactly what all we need.
> > If you need to change response type over interceptor, please consider using a `sum type` with `oneof` keyword.
> 
> @esenmx can you elaborate on the `wrap` extension approach? Let's take one simple use case like measuring the time it takes for `invoker(method, request, _inject(options))` to finish in `interceptUnary`. No response type changed.

If you are talking about merging the `timeout` value or any other metric info into your response data, you should use `trailers` instead. It's why made for, no type changed.

```dart
@override
ResponseFuture<R> interceptUnary<Q, R>(
    ClientMethod<Q, R> method, Q request, CallOptions options, ClientUnaryInvoker<Q, R> invoker) {
  final stopwatch = Stopwatch()..start();
  return invoker(method, request, options)
    ..trailers.then((trailers) {
      trailers['timeout'] = stopwatch.elapsedMilliseconds.toString();
    });
}

// request
final result = client.doSomething();
result.trailers.then((value) {
  print('timeout: ${value['timeout']}');
});
```

esenmx:
> > The response type shouldn't be changed over interceptor, it's obviously very bad idea, on top of that we lose type-safety.
> > Interceptors should behave as complementary like `refreshing tokens`, then `retrying the failed request`, which requires an `async` function body.
> > Please just let `wrap` extension method with `type-safety` that's exactly what all we need.
> > If you need to change response type over interceptor, please consider using a `sum type` with `oneof` keyword.
> 
> @esenmx can you elaborate on the `wrap` extension approach? Let's take one simple use case like measuring the time it takes for `invoker(method, request, _inject(options))` to finish in `interceptUnary`. No response type changed.

If you are talking about merging the `timeout` value or any other metric info into your response data, you should use `trailers` instead. It's why made for, no type changed.

```dart
@override
ResponseFuture<R> interceptUnary<Q, R>(
    ClientMethod<Q, R> method, Q request, CallOptions options, ClientUnaryInvoker<Q, R> invoker) {
  final stopwatch = Stopwatch()..start();
  return invoker(method, request, options)
    ..trailers.then((trailers) {
      trailers['timeout'] = stopwatch.elapsedMilliseconds.toString();
    });
}

// request
final result = client.doSomething();
result.trailers.then((value) {
  print('timeout: ${value['timeout']}');
});
```

lib/src/client/common.dart
acoutts: One idea- could you make an extension on Future / ResponseFuture instead?

mraleph:

Let me know if that would solve all of your potential use cases.

mraleph:
I am going to close this PR because direct chaining has some hidden pitfalls (e.g. one highlighted in https://github.com/grpc/grpc-dart/pull/419#issuecomment-854542925).

PR #423, opened 2021-01-15
Add XHR raw response to the GrpcError for a better debugging

mightyvoice:
Add XHR raw response to the GrpcError for a better debugging.


mraleph (approved):
LGTM 

Though I am bit concerned that this is only tested with a mock connection and not an actual browser XHR - which might have some different lifecycle behavior.

mightyvoice:
> LGTM
> 
> Though I am bit concerned that this is only tested with a mock connection and not an actual browser XHR - which might have some different lifecycle behavior.

It has been used internally in Google by many teams for a while without any related bug related.

PR #427, opened 2021-01-22
Make sure that grpc_web.dart builds on the Web

mraleph:
c48af638a54225fe39200df726440ac7bde4225c accidentally introduced a dependency on VM specific libraries in the shared code (apparently not caught by the tests?). `webdev` however refuses to build `grpc_web.dart` after it.


mraleph:
TBR @sigurdm

sigurdm (commented):
LGTM

PR #435, opened 2021-01-27
Fix potential NPE

ankur2136:


mraleph (approved):
Normally I'd ask to make a test, but given NNBD migration I would ignore this :)

PR #436, opened 2021-01-28
Proper NPE fix

ankur2136:
Sorry for the past PR which didn't completely fix the NPE in this code.

https://stackoverflow.com/a/51141153/2531686


mraleph (approved):

mraleph:
(And that's why adding tests is always good :))

PR #454, opened 2021-03-07
Avoid completing multiple times for one error

kevmoo:
Also:
* updated SDK constraint to stable
* Use latest Dart action install
* Test on oldest supported SDK

### DIFF
--- a/interop/lib/src/client.dart
+++ b/interop/lib/src/client.dart
@@ -1108,12 +1108,14 @@ class Tester {
       }
       call.cancel();
     }, onError: (e) {
-      if (e is! GrpcError) completer.completeError('Unexpected error: $e.');
-      if (e.code != StatusCode.cancelled) {
+      if (e is! GrpcError) {
+        completer.completeError('Unexpected error: $e.');
+      } else if (e.code != StatusCode.cancelled) {
         completer
             .completeError('Unexpected status code ${e.code}: ${e.message}.');
+      } else {
+        completer.complete(true);
       }
-      completer.complete(true);
     }, onDone: () {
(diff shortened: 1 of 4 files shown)


interop/lib/src/client.dart
```diff
@@ -1108,12 +1108,14 @@ class Tester {
       }
       call.cancel();
     }, onError: (e) {
-      if (e is! GrpcError) completer.completeError('Unexpected error: $e.');
-      if (e.code != StatusCode.cancelled) {
+      if (e is! GrpcError) {
+        completer.completeError('Unexpected error: $e.');
```
mraleph, line 1112: It would really be much simpler to read if this logic was just written using an `async` function. Something like this:

```dart
try {
  var receivedResponse = false;
  await for (var response in call) {
    if (receivedResponse) {
      throw 'Received too many responses';
    }
    if (response.payload.body.length != 31415) {
      completer.completeError('Invalid response length: '
          '${response.payload.body.length} != 31415.');
    }
    call.cancel();
  }
  throw 'Expected error';
} catch (e) {
  if (e is! GrpcError) {
    // Alternatively just rethrow;
    throw 'Unexpected error: $e.';
  } else if (e.code != StatusCode.cancelled) {
    throw 'Unexpected status code ${e.code}: ${e.message}.';
  }
  // Got a GrpcError with StatusCode.cancelled as expected.
}
```

mraleph (approved):

interop/lib/src/client.dart
kevmoo: Agreed. I was trying to do the smallest delta here.

Issue #421, opened 2021-01-15
Use HTTP code to figure out appropriate GrpcError if no grpc-status header exists


rohitramkumar:
I have a setup where my Dart frontend is calling out to an API fronted by Envoy. The Envoy performs a JWT + RBAC check before forwarding requests to the backend. If the JWT is not valid or the user is denied as per the RBAC config, then Envoy returns a 403 but it does not return the `grpc-status` header. As far as I can tell, the underlying libraries here are specifically looking for this header to figure out what kind of GrpcError to return (which seems reasonable). However, in cases where no header is returned, I feel it should look to the HTTP code to make the decision. Right now, the transport layer simply calls `Grpc.unavailable()` if the HTTP code is not 200 [1]. Since this code seems to be already looking at headers, can we make this logic a bit smarter?

Specifically, I'm thinking that if HTTP code is not 200, then have a large switch statement that maps the HTTP code to the proper GrpcError. Would be happy to send a PR if this sounds reasoanable. FWIW, there seems to be some precedent for a change like this (https://github.com/grpc/grpc-web/pull/619).

[1]  https://github.com/grpc/grpc-dart/blob/master/lib/src/client/transport/xhr_transport.dart#L118

LeFrosch:
I’m currently facing a similar issue. My dart client tries to connect to a server which also uses JWT based authentication. The client uses an API which returns a stream of results but if the authentication fails no error gets thrown at all. The `ResponseStream` is just empty.
I think it would be more intuitive to throw an error in this case instead of returning an empty stream. The responsible code snippet can be found [here](https://stackoverflow.com/questions/66680087/dart-grpc-does-not-throw-an-exception-when-authorization-fails). I’m sorry if this problem is caused by me, I’m quite new to grpc and dart.

mraleph:
Yeah it seems we are not doing enough checking when receiving responses (see also #458 which is caused by lack of checks in the parser that handles native gRPC responses). 

We would be happy to take PRs that fix this.

PR #460, opened 2021-03-18
Be more resilient to broken deployments

LeFrosch:
If there is no `grpc-status` found in the header, I try to convert the http status into the corresponding grpc status. For the conversion I used the `fromHttpStatus` function from [this](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/statuscode.js) as a template.
This fixes the issue I had here #421


lib/src/client/call.dart
```diff
@@ -346,7 +346,17 @@ class ClientCall<Q, R> implements Response {
...
-    final statusCode = int.parse(status ?? '0');
+    late final int statusCode;
+
+    if (status != null) {
+      statusCode = int.parse(status);
+    } else {
+      final httpStatus = metadata[':status'];
+
+      statusCode = httpStatus == null
+          ? 0
```
mraleph, line 357: I think this can be `StatusCode.unknown` or `StatusCode.internal` - if response does not contain either `:status` or `grpc-status` then something is serious wrong with it.

lib/src/shared/status.dart
```diff
@@ -120,6 +120,43 @@ class StatusCode {
   /// The request does not have valid authentication credentials for the
   /// operation.
   static const unauthenticated = 16;
+
+  /// Creates a gRPC Status code from a HTTP Status code
+  ///
+  /// Copied from: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/statuscode.js
+  static int fromHttpStatus(int status) {
```
mraleph, line 127: I suggest using a map here: 

```dart
const statusMapping = {200: Status.ok, 400: StatusCode.invalidArgument, ...};

return statusMapping[status] ?? StatusCode.unknown;
```

Even better for readability would be if you do:

```dart
// It is okay to import dart:io into shared code as long as you don't use
// any runtime functionality.
import 'dart:io' show HttpStatus;

const _httpStatusToGrpcStatus = <int, int>{
  HttpStatus.ok: StatusCode.ok,
  HttpStatus.badRequest: StatusCode.invalidArgument,
  HttpStatus.unauthorized: StatusCode.unauthenticated,
  HttpStatus.forbidden: StatusCode.permissionDenied,
  HttpStatus.notFound: StatusCode.notFound,
  HttpStatus.conflict: StatusCode.aborted,
  HttpStatus.preconditionFailed: StatusCode.failedPrecondition,
  HttpStatus.tooManyRequests: StatusCode.resourceExhausted,
  HttpStatus.clientClosedRequest: StatusCode.cancelled,
  HttpStatus.internalServerError: StatusCode.unknown,
  HttpStatus.notImplemented: StatusCode.unimplemented,
  HttpStatus.serviceUnavailable: StatusCode.unavailable,
  HttpStatus.gatewayTimeout: StatusCode.deadlineExceeded,
};
```

mraleph (changes requested):
This looks good. Thank you for taking time to make a PR. 

I have made few suggestions how to make it a bit more robust. 

Please consider also adding at least few tests to make sure we don't regress this, but it's also okay to skip that.

lib/src/client/call.dart
```diff
@@ -346,7 +346,17 @@ class ClientCall<Q, R> implements Response {
   /// If there's an error status then process it as a response error
   void _checkForErrorStatus(Map<String, String> metadata) {
     final status = metadata['grpc-status'];
-    final statusCode = int.parse(status ?? '0');
+    late final int statusCode;
+
+    if (status != null) {
+      statusCode = int.parse(status);
+    } else {
+      final httpStatus = metadata[':status'];
```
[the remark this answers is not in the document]
LeFrosch: The first line doesn’t compile with null safety, because `metadata.containsKey(':status')` does not ensure that `metadata[':status']` cannot return null and `int.tryParse` expects a not nullable String. I’ll just tweak it a little bit to make it work with null safety.
But I noticed the header validation breaks a lot of tests, because most of them don’t populate the headers with the `status` and the `content-type`. In addition I tried the header validation with our asp .net core server and the server doesn’t seem to send the content type, thus this change would break our application. 
Personally I would suggest to only check if the http code is present when validating the header.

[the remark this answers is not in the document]
mraleph: > In addition I tried the header validation with our asp .net core server and the server doesn’t seem to send the content type, thus this change would break our application.

Interesting. That's a violation of gRPC specifications. I see that https://github.com/grpc/grpc-dotnet seems to send back correct Content-Type, are you using some custom gRPC implementation for .NET? I also see that most native clients validate Content-Type they receive.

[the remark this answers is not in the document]
mraleph: I also looked at grpc-web implementation and it does validate content type it receives from the server: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/grpcwebclientreadablestream.js#L138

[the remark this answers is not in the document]
LeFrosch: My bad, it was because of something else that the content-type was not sent in my case. Sorry for wasting your time.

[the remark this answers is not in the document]
mraleph: No worries! You are *not* wasting my time in any way. Thank *you* for spending time on this PR, this is really appreciated.

test/client_tests/client_test.dart
```diff
@@ -173,7 +173,7 @@ void main() {
 
     await harness.runFailureTest(
       clientCall: harness.client.unary(dummyValue),
-      expectedException: GrpcError.unimplemented('No responses received'),
+      expectedException: GrpcError.unknown('missing :status header'),
```
LeFrosch, line 176: I'm not quite sure if this is the way to go, because the former exception seems to fit better?

test/src/client_utils.dart
```diff
@@ -233,7 +233,11 @@ abstract class _Harness {
     toClient.close();
   }
 
-  void sendResponseHeader({List<Header> headers = const []}) {
+  void sendResponseHeader() {
+    final headers = [
+      Header.ascii(':status', '200'),
+      Header.ascii('content-type', 'application/grpc'),
+    ];
```
LeFrosch, line 240: I noticed that the function arguments were never used. But I’m also not sure if this is the right way to go.

test/client_tests/client_test.dart
mraleph: This can be fixed by sending `:status` as part of the trailers. gRPC protocol assumes that OnlyTrailers responses contain both trailers and headers. 

I am going to push a commit on top of your work fixing this.

test/src/client_utils.dart
mraleph: Yeah, I think this is good fix.

mraleph:
@LeFrosch okay, so I have spent some time today figuring how the best structure these checks. I think I have arrived to a good approach at the end and pushed some changes into your repo. 

But now I have made so many changes myself, that I should not be the one reviewing this code. I am going to ask @mkustermann to take a look on Monday and then once he approves I will land the code.

LeFrosch:
@mraleph Thanks a lot for the help and effort :)

test/grpc_web_test.dart
```diff
@@ -116,14 +116,34 @@ void main() {
...
+    'bad-content-type':
+        GrpcError.unknown('unsupported content-type (text/html)'),
+  };
+
+  for (var entry in invalidResponseTests.entries) {
+    test('invalid response: ${entry.key}', () async {
+      final channel = GrpcWebClientChannel.xhr(server.httpUri);
+      final service = EchoServiceClient(channel,
+          options: WebCallOptions(bypassCorsPreflight: true));
+      expect(() => service.echo(EchoRequest()..message = 'test:${entry.key}'),
```
mkustermann, line 134: Maybe add a comment explaining that the server will treat `EchoRequest` with `test:` specially by returning errors instead of echoing anything.

test/grpc_web_server.dart
```diff
@@ -164,4 +168,54 @@ if you are running tests locally.
     tempDir.deleteSync(recursive: true);
   }
   channel.sink.add('EXITED');
+
+  await server.shutdown().catchError((_) {});
+  await httpServer.close().catchError((_) {});
```
mkustermann, line 173: Are these ignore-all-errors catches necessary here? Could they hide failures in the tests?

test/grpc_web_server.dart
```diff
@@ -164,4 +168,54 @@ if you are running tests locally.
...
+    rs.close();
+  },
+};
+
+void defaultHandler(HttpResponse rs) {
+  rs.close();
+}
+
+Future<HttpServer> startHttpServer() async {
+  final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 51428);
```
mkustermann, line 202: Could you avoid hard-coding the port?

test/client_tests/client_test.dart
```diff
@@ -286,13 +289,86 @@ void main() {
     );
   });
 
+  test('Call throws if HTTP status indicates an error', () async {
+    void handleRequest(_) {
+      harness.toClient.add(HeadersStreamMessage([
+        Header.ascii(':status', HttpStatus.serviceUnavailable.toString()),
+        Header.ascii('content-type', 'application/grpc'),
+      ]));
+      // Send a frame that might be misinterpreted as header frame and cause
```
mkustermann, line 298: I think it cannot be misinterpreted as header frame - http/2 will treat it as data frame(s).

as a header frame -> as a length-prefixed proto message?

(also below)

lib/src/shared/streams.dart
```diff
@@ -117,7 +125,17 @@ class _GrpcMessageConversionSink extends ChunkedConversionSink<StreamMessage> {
...
     }
-    // TODO(jakobr): Check :status, go to error mode if not 2xx.
+    if (!_headersReceived) {
+      if (_forResponse) {
+        // Validate :status and content-type header here synchronously before
+        // attempting to parse subsequent DataStreamMessage.
+        final httpStatus = headers.containsKey(':status')
+            ? int.tryParse(headers[':status']!)
+            : null;
+        validateHttpStatusAndContentType(httpStatus, headers);
```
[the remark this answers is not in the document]
mkustermann: There are two kind of responses:

  * `HttpHeaders` + `Data`... + `HttpHeaders` (tailer) + EOS
  * `HttpHeader` EOS

In the first case (normal case) the http `:status 200` is expected in the first `HttpHeaders` and the `grpc-status` is expected in the trailer `HttpHeaders`. 

It seems this code is handling/looking at both. As discussed offline you mentioned the trailer handling looking only at `grpc-status` is in a different place, consider linking to that.

mkustermann (approved):

lib/src/shared/status.dart
```diff
@@ -309,3 +336,96 @@ GeneratedMessage parseErrorDetailsFromAny(Any any) {
...
+}
+
+GrpcError? grpcErrorFromTrailers(Map<String, String> trailers) {
+  final status = trailers['grpc-status'];
+  final statusCode = status != null ? int.parse(status) : StatusCode.unknown;
+
+  if (statusCode != StatusCode.ok) {
+    final messageMetadata = trailers['grpc-message'];
+    final message =
+        messageMetadata == null ? null : Uri.decodeFull(messageMetadata);
```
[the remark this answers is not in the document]
mraleph: Good point. Specification actually tells us to be lenient and preserve incorrect messages which can't be decoded. Fixed.

test/src/client_utils.dart
```diff
@@ -233,17 +235,29 @@ abstract class _Harness {
...
+    Header.ascii(':status', '200'),
+    Header.ascii('content-type', 'application/grpc'),
+  ];
+
+  static final _defaultTrailers = [
+    Header.ascii('grpc-status', '0'),
+  ];
+
+  void sendResponseHeader() {
+    headersWereSent = true;
```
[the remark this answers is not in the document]
mraleph: Done.

lib/src/shared/status.dart
```diff
@@ -120,6 +126,27 @@ class StatusCode {
   /// The request does not have valid authentication credentials for the
   /// operation.
   static const unauthenticated = 16;
+
+  static const _httpStatusToGrpcStatus = <int, int>{
```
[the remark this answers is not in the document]
mraleph: Done.

test/grpc_web_server.dart
mraleph: Removed `catchError`.

test/grpc_web_server.dart
mraleph: Good catch. This was left-over from debugging.

test/grpc_web_test.dart
mraleph: Done.

lib/src/shared/streams.dart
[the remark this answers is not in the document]
mraleph: > If this method throws, which code will catch it and report it to user code?

The `StreamTransformer.bind` implementation (or more specifically `_SinkTransformerStreamSubscription`) takes care of catching exceptions that occur in transformer and forwarding it into the stream as an error. Added a comment explaining this. 

> It seems this code is handling/looking at both. As discussed offline you mentioned the trailer handling looking only at grpc-status is in a different place, consider linking to that.

Comment added.

mraleph (approved):

mraleph:
Thanks for taking a look @mkustermann. Waiting for tests to go green and then will land.

PR #472, opened 2021-04-30
Make clientCertificate available in ServiceCall

EPNW:
The current server implementation does not allow to access the certificate of the client in an incoming call. The main reason for accessing the client certificate in a call is to authenticate the underlying user. To this purpose we extended the ServiceCall with a new getter `X509Certificate? get clientCertificate;`. The client certificate is then obtained from the SecureSocket and passed through. To request/require the client certificate, the `Server.serve()` method has two new optional parameters `requestClientCertificate` and `requireClientCertificate`, which are used in `SecureServerSocket.bind()` to make the certificate request.

### DIFF
--- a/test/client_certificate_test.dart
+++ b/test/client_certificate_test.dart
@@ -1,4 +1,6 @@
-@TestOn('vm')
+// TODO(dartbug.com/26057) currently Mac OS X seems to have some issues with
+// client certificates so we disable the test.
+@TestOn('vm && !mac-os')
 import 'dart:async';
 import 'dart:io';
 
--- a/test/client_certificate_test.dart
+++ b/test/client_certificate_test.dart
@@ -23,83 +23,74 @@ class EchoService extends EchoServiceBase {
 }
 
 const String address = 'localhost';
-const int port = 19360;
 Future<void> main() async {
+  setUp(() {});
   test('Client certificate required', () async {
-    //Server
-    final server = Server(
-      [EchoService()],
-    );
-    final serverContext =
-        SecurityContextChannelCredentials.baseSecurityContext();
+    // Server
+    final server = await _setUpServer(true);
 
-    serverContext.useCertificateChain('test/data/localhost.crt');
-    serverContext.usePrivateKey('test/data/localhost.key');
-    serverContext.setTrustedCertificates('test/data/localhost.crt');
-    final ServerCredentials serverCredentials =
-        SecurityContextServerCredentials(serverContext);
-
-    await server.serve(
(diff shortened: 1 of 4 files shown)


mraleph:
Hi! Thank you for the contribution. Please add tests for the added functionality.

EPNW:
Provided test, see fd19f01fc5311051c0ab2eef177b7966f5868a83.

test/client_certificate_test.dart
```diff
@@ -0,0 +1,131 @@
...
+  @override
+  Stream<ServerStreamingEchoResponse> serverStreamingEcho(
+      ServiceCall call, ServerStreamingEchoRequest request) {
+    // TODO: implement serverStreamingEcho
+    throw UnimplementedError();
+  }
+}
+
+const String address = 'localhost';
+const int port = 19360;
```
mraleph, line 26: You should avoid hardcoding the port, because the test can easily fail if the port is not available. Instead you can ask `server.port` after you start the server.

test/client_certificate_test.dart
```diff
@@ -0,0 +1,131 @@
...
+Future<void> main() async {
+  test('Client certificate required', () async {
+    //Server
```
mraleph, line 29: `// Server` (whitespace after `// `).

test/client_certificate_test.dart
```diff
@@ -0,0 +1,131 @@
...
+        port: port, options: ChannelOptions(credentials: channelCredentials));
+    final client = EchoServiceClient(channel);
+    expect((await client.echo(EchoRequest())).message, '/CN=localhost');
+    await channel.shutdown();
+    await server.shutdown();
+  });
+
+  test('Client certificate not required', () async {
+    //Server
+    final server = Server(
```
mraleph, line 69: Consider moving `server` setup into a helper method because it is duplicated between two tests.

mraleph (changes requested):

EPNW:
Implemented requested changes, see 08128cf92a168b2f846c968dcf24cbb171693b0c and cc44119330eb176110237d8bd531246253a24fd0.

mraleph:
I have pushed a minor change (`port: 0`) to get OS to select the port (in my previous comment I was assuming that not passing the port leads to this behaviour - but it turns out that `serve` would then default to 80 or 443 instead of letting OS select the port). 
 
Unfortunately there seem to be some problem with Mac OS X client certificate handling. I am going to simply suppress this test on Mac OS X for now just to allow this PR to land.

Issue #482, opened 2021-06-03
GrpcOrGrpcWebClientChannel   grpcport and grpcWebPort  must not be prevented to be the same.


gustav3d:
I'ts false assumption that there need to be two different ports, and preventing to use the same port makes it not work with standard configuration for GRPC AspNetCore server.

mraleph:
Feel free to submit a PR.

gustav3d:
Feel free to fix your own flawed design. Dont expect others to do it for you.

mraleph:
I understand you frustration, but you should understand that this repo is not really actively maintained. I am reviewing PRs and keeping things somewhat afloat more or less out of sense of responsibility. Protocols are also rather remote from my area of expertise (which is compiler contruction), so understandably I have no way of knowing that there are gRPC servers that serve both grpc and grpc-web protocols on the same port. 

Hence if you want something to be fixed, you shortest path is to fix it yourself. 

Alternatively, others (maybe @morgwai who contributed the original code) can fix it. So lets keep it open.

morgwai:
I wasn't aware about this either: all the docs I've seen were clearly stating that these are separate protocols. I assumed that this was not possible due to the fact that gRPC does not seem to have a notion of request-path prefix.
@gustav3d  Could you explain or link to some docs that explain how this is achieved? I'd be eager to understand it ;-)

A fix for this is quite trivial: remove 1 line that does the check. Actually it's 2 lines as there are 2 versions for web platform and for others ;-)
https://github.com/grpc/grpc-dart/blob/master/lib/src/client/grpc_or_grpcweb_channel_grpc.dart#L36
https://github.com/grpc/grpc-dart/blob/master/lib/src/client/grpc_or_grpcweb_channel_web.dart#L28

morgwai:
draft: https://github.com/grpc/grpc-dart/pull/483

mraleph:
I'd guess that middleware can detect protocol type by peaking at `Content-Type` header, which is different for these two protocols.

mraleph:
Fixed by @morgwai ! Thanks!

morgwai:
@mraleph thanks for the review!
@gustav3d  thanks for pointing the issue!

PR #483, opened 2021-06-03
GrpcOrGrpcWeb: remove checking if ports are different (#482)

morgwai:
Using the same port is a standard for AspNetCore gRPC server.

@mraleph
@gustav3d

### DIFF
--- a/lib/grpc_or_grpcweb.dart
+++ b/lib/grpc_or_grpcweb.dart
@@ -17,9 +17,13 @@
 /// on all platfroms except web, on which it uses [GrpcWebClientChannel].
 ///
 /// Note that gRPC and gRPC-web are 2 different protocols and server must be
-/// able to speak both of them (usually on separate ports) for this to work.
-/// Therefore applications using this class must provide both ports and the
-/// channel will use the one for the actual protocol being used.
+/// able to speak both of them for this to work.
+/// As several existing implementations (such as in-process gRPC-web to gRPC
+/// proxies or Envoy gRPC-web to gRPC proxy) expose gRPC and gRPC-web on
+/// separate ports, the constructor requires 2 ports to be provided and
+/// the channel will use the one for the actual protocol being used.
+/// If the server supports both protocols on the same port (such as AspNetCore
+/// implementation), then the same port value should be provided on both params.
 
 export 'src/client/grpc_or_grpcweb_channel_grpc.dart'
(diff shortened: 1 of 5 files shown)


lib/grpc_or_grpcweb.dart
```diff
@@ -17,9 +17,9 @@
 /// on all platfroms except web, on which it uses [GrpcWebClientChannel].
 ///
 /// Note that gRPC and gRPC-web are 2 different protocols and server must be
```
mraleph, line 19: Consider removing "usually on separate ports". Maybe: 

```dart
// We allow to specify two different ports in case
// if the same port does not speak both protocols. 
```

lib/src/client/grpc_or_grpcweb_channel_web.dart
```diff
@@ -23,11 +23,7 @@ class GrpcOrGrpcWebClientChannel extends GrpcWebClientChannel {
     required int grpcWebPort,
```
mraleph, line 23: Maybe we should consider changing this constructor to simply be: 

```dart
GrpcOrGrpcWebClientChannel.toSeparateEndpoints({
  required String grpcHost,
  required int grpcPort,
  required String grpcWebHost,
  required int grpcWebPort,
}) ...;

GrpcOrGrpcWebClientChannel({
  required String host,
  required int port
}) : this.toSeparateEndpoints(...);  // Pass the same host and port.
```

Then if server can serve both then you just golden, otherwise you use `toSeparateEndpoints` (and we allow host to be configurable as well). 

WDYT?

mraleph (changes requested):

lib/grpc_or_grpcweb.dart
morgwai: I've reworded this source doc completely: please have a look.

lib/src/client/grpc_or_grpcweb_channel_web.dart
morgwai: Your solution is more flexible in general, but I also think that same-host+2-different-ports is the most common setup, so it would probably be nice to provide constructor for it, I think.

Nevertheless, I think this should not be a part of this PR, which basically aims to fix #482
I've created a [separate issue for this](https://github.com/grpc/grpc-dart/issues/484) and will prepare some draft PR (tomorrow probably) so we can discuss it there.

mraleph (approved):

Issue #484, opened 2021-06-03
consider providing more flexible constructors for GrpcOrGrpcWebClientChannel


morgwai:
Current constructors assume the same host for both gRPC and gRPC-web: this may not be flexible enough and may require complicating application code: library should probably make it easier for developers.
See https://github.com/grpc/grpc-dart/pull/483#discussion_r644644157

PR #485, opened 2021-06-03
GrpcOrGrpcWeb: provide more flexible constructors (#484)

morgwai:
@mraleph 

btw: it would be soooo awesome if https://github.com/dart-lang/language/issues/723 was implemented, so that it would be transparent for library users if they are using a 'native' constructor or an extension  ;-)
(for example [here](https://github.com/morgwai/grpc-dart/blob/grpcOrWebConstructors/test/client_tests/grpc_or_grpcweb_channel_grpc_test.dart#L26) I could write
```dart
GrpcOrGrpcWebClientChannel.toSingleEndpoint(...);
```
instead of 
```dart
GrpcOrGrpcWebClientChannelConstructors.toSingleEndpoint(...);
```
)

### DIFF
--- a/lib/grpc_or_grpcweb.dart
+++ b/lib/grpc_or_grpcweb.dart
@@ -16,10 +16,10 @@
 import 'src/client/grpc_or_grpcweb_channel_grpc.dart'
     if (dart.library.html) 'src/client/grpc_or_grpcweb_channel_web.dart';
 import 'src/client/http2_channel.dart';
+import 'src/client/options.dart';
 
-/// Exports [GrpcOrGrpcWebClientChannel] that underneath uses gRPC
-/// [ClientChannel] on all platforms except web, on which it uses
-/// [GrpcWebClientChannel].
+/// A client channel that underneath uses gRPC [ClientChannel] on all platforms
+/// except web, on which it uses [GrpcWebClientChannel].
 ///
 /// Note that gRPC and gRPC-web are 2 different protocols and server must be
 /// able to speak both of them for this to work.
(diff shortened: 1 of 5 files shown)


mraleph (approved):

mraleph:
On the second though... I am not sure this `GrpcOrGrpcWebClientChannelConstructors.toSingleEndpoint` looks nice. Any reason you would not make it a constructor on `GrpcOrGrpcWebClientChannel` itself? To avoid too much duplication you could rename `GrpcOrGrpcWebClientChannel` into `GrpcOrGrpcWebClientChannelImpl` and then create `GrpcOrGrpcWebClientChannel` which extends `GrpcOrGrpcWebClientChannelImpl` and delegates to the appropriate constructors.

morgwai:
> On the second though... I am not sure this `GrpcOrGrpcWebClientChannelConstructors.toSingleEndpoint` looks nice. 

it is not beautiful, to say least ;-] that's why I'm whining about https://github.com/dart-lang/language/issues/723 ;-)

> Any reason you would not make it a constructor on `GrpcOrGrpcWebClientChannel` itself? To avoid too much duplication you could rename `GrpcOrGrpcWebClientChannel` into `GrpcOrGrpcWebClientChannelImpl` and then create `GrpcOrGrpcWebClientChannel` which extends `GrpcOrGrpcWebClientChannelImpl` and delegates to the appropriate constructors.

so basically replace using extension with using inheritance: yeah, I will change it this way.

morgwai:
@mraleph  ready: please have another look in a spare moment )

mraleph (approved):

PR #492, opened 2021-06-27
Fix a typo on the inner invoker parameters

balesz:
The operations on the request stream of the ClientInerceptor items are ignored when calls the last invoker which creates the ResponseStream.


mraleph:
Thanks for the fix! Would you mind adding a regression test as well?

balesz:
@mraleph I added a new test case for the fix.

mraleph (approved):

mraleph:
Thanks a lot! CI is running - I will merge when it cycles green.

PR #493, opened 2021-06-28
Add custom trailers to GrpcError

### LINKED ISSUE #135: Feature request: expose trailers from GrpcError
Until now, it seems GrpcError can only receive the normal error. But grpc-java can receive trailers from StatusRuntimeException just like class GrpcError.

I need to get the trailers from GrpcError to deal business logic.

version: 0.6.6

kengu:
This PR resolves #135.

**Background**
Servers can return additional error information as `*Custom-Metadata` in `Trailers`. These are not exposed in `GrpcError` making it practically impossible to implement typed exception handling in `ClientInterceptor`s. Exposing trailers in error objects is supported in [other implementations like csharp](https://grpc.github.io/grpc/csharp/api/Grpc.Core.RpcException.html#Grpc_Core_RpcException_Trailers), and [examples in the wild ](https://github.com/EventStore/EventStore-Client-Dotnet/blob/a853663cb06fcc70db12f0da20956f4afa63f3c7/src/EventStore.Client/Interceptors/TypedExceptionInterceptor.cs#L99) depends on custom trailers to communicate error information. 

**Tests**
A new test "Call should throw with custom trailers" is added to verify that custom headers only are exposed.

### DIFF
--- a/lib/src/shared/status.dart
+++ b/lib/src/shared/status.dart
@@ -160,7 +160,7 @@ class GrpcError implements Exception {
 
   /// Custom error code.
   GrpcError.custom(this.code,
-      {this.message, this.details, this.rawResponse, this.trailers = const {}});
+      [this.message, this.details, this.rawResponse, this.trailers = const {}]);
 
   /// The operation completed successfully.
   GrpcError.ok([this.message, this.details, this.rawResponse])
@@ -399,11 +399,11 @@ void validateHttpStatusAndContentType(
     if (error == null || error.code == StatusCode.unknown) {
       throw GrpcError.custom(
         status,
-        message: error?.message ??
+        error?.message ??
             'HTTP connection completed with ${httpStatus} instead of 200',
-        details: error?.details,
-        rawResponse: rawResponse,
-        trailers: error?.trailers ?? toCustomTrailers(headers),
+        error?.details,
+        rawResponse,
+        error?.trailers ?? toCustomTrailers(headers),
       );
     }
     throw error;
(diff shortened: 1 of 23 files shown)


kengu:
@ejona86 / @mit-mit This is my first PR to grcp-dart, so I hope I've followed steps required for a review.

kengu:
@mraleph Do you have time to look at this PR?

mraleph:
@kengu I will try to find some time, maybe next week.

kengu:
> @kengu I will try to find some time, maybe next week.

Hi @mraleph, I'm currently testing this branch with a git-ref i pubsec.yaml, and it works as intended. I hope you find some time to make a review :pray:

lib/src/shared/status.dart
```diff
@@ -155,62 +155,73 @@ class GrpcError implements Exception {
   final int code;
   final String? message;
   final Object? rawResponse;
+  final Map<String, String>? trailers;
   final List<GeneratedMessage>? details;
 
   /// Custom error code.
-  GrpcError.custom(this.code, [this.message, this.details, this.rawResponse]);
```
mraleph, line 161: This is a breaking change in a public constructor. Could we avoid it?

mraleph (changes requested):
I think this LGTM, though I would prefer to avoid breaking change to `GrpcError.custom` API.

lib/src/shared/status.dart
kengu: Agree, I'll refactor.

mraleph (approved):

kengu:
@mraleph I see that dart format checks fails now. Running `dart format --output=none --set-exit-if-changed .` locally does not fail.

mraleph:
Trying to address it here: https://github.com/grpc/grpc-dart/pull/504

mraleph:
d'oh. Now there is another problem with `envoy` package not available anymore through https://dl.bintray.com/tetrate/getenvoy-deb - unfortunately I don't think I will have enough time left in the day to fix this today. I will pick this up when I am back from sommerferien

mraleph:
@kengu could you try rebasing on top of master? I think CI should be green now.

mraleph:
Thanks for the PR. Merged.

kengu:
@mraleph Do you know when the next release of grpc-dart to pub.dev is scheduled? I still need this before I can publish my own packaged that depends on this change.

mraleph:
Released 3.0.1

PR #494, opened 2021-06-29
Fix interop/bin/server.dart after NNBD migration

cyrax111:


mraleph:
Thanks for the fix. Could you fix CLA problem so that we could merge?

cyrax111:
I will try ) Do I need to make new commit and pull request after GitHub email changing ?

mraleph:
I think you should be able to force push into the same branch to completely overwrite the history once you fix email on the commit.

cyrax111:
CLA problem was fixed.

mraleph:
Thanks for the fix. Applied.

PR #504, opened 2021-07-16
Reformat code with dev build of dartfmt

mraleph:


mraleph:
Reference for a future self: envoy installation is failing because https://github.com/envoyproxy/envoy/issues/17236

PR #512, opened 2021-08-04
Add Ctrl+C Handler

stellarpower:
If using Unix domain sockets, the old socket needs to be cleaned up properly, or restarting the server will cause it to refuse to bind. Added a handler for SIGINT to call shutdown explicitly and then exit.

This is the first time I've written in Dart - am completely open to changes.

### DIFF
--- a/example/helloworld/bin/server.dart
+++ b/example/helloworld/bin/server.dart
@@ -33,4 +33,11 @@ Future<void> main(List<String> args) async {
   );
   await server.serve(port: 50051);
   print('Server listening on port ${server.port}...');
+  ProcessSignal.SIGINT.watch().listen((signal) {
+    if (signal == ProcessSignal.SIGINT){
+      print("Shutting down");
+      server.shutdown();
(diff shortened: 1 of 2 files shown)


mraleph:
Thanks for taking your time to contribute. 

I think a better alternative is just to `unlink` (e.g. `File.deleteSync`) UDS path before trying to bind. Also you need to sign the CLA before I can look at/merge the code.

stellarpower:
I agree that makes sense, but I think also gracefully cleaning up is important anyway. IMO should be encouraged as that's the general solution to the problem. I leave it up to you, I ended up abandoning the efforts in Dart and trying Go. It was interesting as a language, and the way it compiles but targets browsers is certainly fantastic, but I don't like Futures - I suppose to kill JS one has to be a bit like JS. Came to the same problem and implemented the same signal handler in Go basically. Wasn't too keen on that as a language, or protocol buffers for that matter, so wasn't Google's week!

PR #516, opened 2021-08-16
Fix compilation with DDC

mraleph:
b272632450498f1c688fe4b0988fa595a0004124 exposed client certificate through `X509Certificate? get clientCertificate;` getter on `ServiceCall` class. This broke compilation of `grpc_web` code using DDC, but not dart2js. Turns out that dart2js is happy to compile any code using `dart:io` (though the result will not run if you try to use any of those APIs), but DDC rejects such code eagerly. `package:test` runs tests through `dart2js` so DDC breakage was not really caught by CI. 

Unfortunately this discrepancy between DDC and dart2js puts us in some really weird spot:  most of our tests are platform independent, but most of those tests also pull in `dart:io` through transitive dependencies.

I tried to do the most minimal change I could to make code compile both on the Web and natively. 

A proper fix should be to go through tests one-by-one and make sure that those that need to run on the Web don't import `dart:io`, but I don't have time to do that right now. 

I have added a smoke test to the CI to verify that `grpc_web` example builds with DDC.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -26,7 +26,8 @@ jobs:
         run: dart --version
       - name: Install dependencies
         run: dart pub get
-      - name: Check formatting
+      - name: Check formatting (using stable dartfmt release)
+        if: ${{ matrix.sdk == 'stable' }}
         run: dart format --output=none --set-exit-if-changed .
       - name: Analyze code (introp and examples)
         run: |
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -41,6 +41,7 @@ jobs:
       - name: Analyze code
         run: dart analyze --fatal-infos .
       - name: Check that grpc-web sample builds with DDC
+        if: ${{ matrix.sdk == 'dev' }}
         run: |
           dart pub global activate webdev
           pushd example/grpc-web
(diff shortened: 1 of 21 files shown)


.github/workflows/dart.yml
```diff
@@ -39,6 +40,15 @@ jobs:
         shell: bash
       - name: Analyze code
         run: dart analyze --fatal-infos .
+      - name: Check that grpc-web sample builds with DDC
```
mkustermann, line 43: Not so familiar with those `dart.yml` files, but it seems a little strange to have building an app as part of the `analyze: ` task.

.github/workflows/dart.yml
```diff
@@ -39,6 +40,15 @@ jobs:
...
+        if: ${{ matrix.sdk == 'dev' }}
+        run: |
+          dart pub global activate webdev
+          pushd example/grpc-web
+          rm -rf build
+          dart pub global run webdev build --no-release
```
mkustermann, line 49: Consider leaving a comment that you run `--no-release` to build the DDC app (I suppose?)

mkustermann (approved):

.github/workflows/dart.yml
mraleph: Done.

.github/workflows/dart.yml
mraleph: Yeah, it is a bit strange, but we use compilation as a way to force dependency checking here (similar to running `dart analyze`).

PR #522, opened 2021-09-02
Remove or skip callbacks on old connections

### LINKED ISSUE #521: Open connections grow when connectionTimeout is used
When connectionTimeout is set (including the default of 50 minutes) and long running streaming requests are run, the number of established HTTP2 connections will increase be one, first on connectionTimeout triggering and then every so often.

I'm narrowing it down more specifically, but roughly the root cause seems to be that we set the transport.onActiveStateChanged callback on the connection when it's established. There will be a running RPC on the connection, so it's active. At some point we realize we want to recycle the connection so we set state to idle, make a new connection, and finish the old one, state becomes ready. Then, the RPC is cancelled on the old connection, it makes the callback with active=false, state gets set to idle, and we reconnect again. We now have two connections. Next interval this repeats, and we get three connections, and so on.

Probably the callback should be removed on connections that get intentionally recycled.

Version 3.0.0

## Repro steps

For easier repro, set a short connectionTimeout.

```
ClientChannel(
      host,
      port: port,
      options: ChannelOptions(
        credentials: ChannelCredentials.secure(certificates: caCertificate),
        connectionTimeout: Duration(minutes: 5),
        idleTimeout: Duration(minutes: 2),
        userAgent: ua,
      ),
    );
```

Run repeated long streaming RPC calls towards some server.

Observe tcp connections using netstat.

Expected result: One connection maintained, recycled at the specified interval.

Actual result: Every interval we gain a new connection.

calmh:
We are not interested in what happens to it after we recycle it, and we don't want the callback to reset our state to idle just after establishing a new connection.

Fixes #521.

Also fixes a small logic error in abandonConnection.


mraleph (approved):

gnossen:
@mraleph 

It looks like this failure [is causing the interop tests to fail](https://source.cloud.google.com/results/invocations/7963b9e1-aadb-40ee-94f6-f1d7ffd3c9a2/targets):

```
2022-12-12 10:07:21,038 TIMEOUT: cloud_to_prod:dart:default:cancel_after_first_response:tls [pid=75134, time=183.6sec]
2022-12-12 10:07:23,167 WAITING: ETA 0.4 sec; 0 queued, 1 jobs running, 441 complete, 1 failed (load 1.00)
2022-12-12 10:07:33,168 WAITING: ETA 0.5 sec; 0 queued, 1 jobs running, 441 complete, 1 failed (load 1.00)
2022-12-12 10:08:23,170 Passed.

2022-12-12 10:08:23,170 TIMEOUT: cloud_to_prod:dart:gateway_v4:cancel_after_first_response:tls [pid=100719, time=189.0sec]
```

PR #543, opened 2022-03-21
feat: allow to make future chain request with ResponseFuture and ResponseStream

savy-91:
Same as https://github.com/grpc/grpc-dart/pull/419 but rebased on master.

Also adds `responseTransform` to chain StreamTransformers to a ResponseStream.


mraleph:

Let me know if that would solve all use cases.

PR #547, opened 2022-04-27
Ensure server's security context is used for the server.

aam:
Use of client's context for the server results in the server not able to correctly handle ALPN selection: server is not able to request http2(the only one that grpc-dart supports) when client offers both http1 and http2, results in client attempting http1 failing to establish the connection.


aam:
@mraleph @kevmoo not sure how to request a review in this project

mraleph:
- If this is solving a particular issue please link to the issue it is solving
- If it is possible to write a regression test in Dart - please include a regression test.

mraleph (approved):

Issue #511, opened 2021-08-04
Memory growth when grpc-web streaming


gawmanarnar:
I currently have an application that streams frames over grpc to flutter mobile and web clients. When running in mobile, the memory seems stable but when doing it in web it appears the data coming from GRPC never gets garbage collected and memory just continues to grow (until the browser runs out of memory). The entirety of these messages are GC'd when the stream is closed but I need to keep the stream open (live video). Is there a way to release these messages as they are processed? I'm wondering if I'm missing something simple. I know grpc-web is done over http but surely this is possible?

Note: the code I wrote is basically the same between platforms and it uses the GrpcOrGrpcWebClientChannel.

Version information:
url: "https://github.com/grpc/grpc-dart.git"
ref: f23070ee85b41504c73cd9586cbb36b383793008
version: "3.0.1-dev"

## Repro steps

1. Have a GRPC server capable of streaming data to a client
2. Create a GrpcOrGrpcWebClientChannel (running on web)
3. Call getStream on it
4. Add a listener to process data

Expected result: Data is garbage collected after being processed

Actual result: Data sticks around and causes memory to grow until out of memory.

## Details

Memory from Mobile (looks reasonable)
![image](https://user-images.githubusercontent.com/2095647/128104138-41f51d4c-1ac9-4bb8-9c90-1d7d51407c11.png)

Memory from Web (constantly increasing)
![image](https://user-images.githubusercontent.com/2095647/128104224-a730bbf1-059d-4334-9bf4-c5bd61c5d210.png)

mraleph:
Unfortunately that's an expected behaviour from the current implementation. 

Currently gRPC-Web protocol implementation builds on top of XHR API which does not actually have any concept of streaming - instead protocol parser simply slices the tail of the XHR response as new bytes arrive (see the code here: https://github.com/grpc/grpc-dart/blob/master/lib/src/client/transport/xhr_transport.dart#L80-L92), which means the browser is slowly accumulating more and more data in memory (the concatenation of all streaming responses).

To address this memory leak one would need to reimplement gRPC-Web transport on top of a streaming friendly Fetch API (or WebSocket API - though WS have their own problems as far as I know).

jbrownsw:
[...]

PR #549, opened 2022-05-04
Implement fetch based transport

jbrownsw:
This is essentially to resolve issue #511. It's designed to be very similar to the existing xhr transport, but with my own `FetchHttpRequest` to replace the `HttpRequest` type. I was a little lazy on the events so they differ a bit, but I could change them to match those used by xhr to make them even more similar. I could also remove some of the code duplication and the transport stream and connection could be shared with only FetchHttpRequest or HttpRequest being the difference and that could be passed in as a function or a generic type parameter. Finally the last thing I wasn't sure of is if you would want this as a complete replacement for xhr or if you'd rather them coexist so people can switch between them etc. So web_channel.dart will need to be updated based on what route you want to go there.

I'm already using this as a drop in replacement for our app and haven't noticed any differences other than the memory growth for streams disappearing.

### DIFF
--- a/lib/src/client/transport/fetch_transport.dart
+++ b/lib/src/client/transport/fetch_transport.dart
@@ -33,6 +33,19 @@ import 'web_streams.dart';
 
 const _contentTypeKey = 'Content-Type';
 
+@JS()
+class AbortSignal {
+  external factory AbortSignal();
+  external bool get aborted;
+}
+
+@JS()
+class AbortController {
+  external factory AbortController();
+  external void abort([dynamic reason]);
+  external AbortSignal get signal;
+}
+
 @anonymous
 // ignore: missing_js_lib_annotation
 @JS()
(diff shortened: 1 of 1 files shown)


jbrownsw:
I created an issue for the dart SDK with my last change (5a8818b):
https://github.com/dart-lang/sdk/issues/49399

listepo:
Hi @jbrownsw, do you think [webfetch](https://github.com/medz/webfetch) could help resolve the issue?

matt-deboer:
@jbrownsw I just tested this branch today (had to add one no-op implementation for the updated main branch) and it seems to work well in debug and release mode. Perhaps your original issues have been since resolved? Or maybe I'm not hitting the repro condition?

jbrownsw:
I don't work at the company that I developed this for so I can't test it, but perhaps @gawmanarnar can get someone to confirm.

PR #351, opened 2020-09-17
Revert "Support grpc-web in pure dart (#287)"

mraleph:
This reverts commit c513e1467f21e498385f4aefa7e370601263034e.

The original commit has broken streaming due to limitations of package:http.

PR #287, opened 2020-03-24
Support grpc-web in pure dart

robsonmeemo:
Fixes #216

Issue #216, opened 2019-07-12
Enable grpc-web without dart:html


sigurdm:
No fundamental reason to limit the grpc-web support to only web-apps as noted by: https://github.com/grpc/grpc-dart/issues/43#issuecomment-510437952 .

joeblew99:
I came here to raise this as an Issue and saw that its already here.

@sigurdm Can you explain by what you mean "without dart:html".

I am wanting to expose a golang Server API to both Flutter and Flutter Web.
https://github.com/grpc-ecosystem/grpc-gateway does not support Web sockets

https://github.com/improbable-eng/grpc-web does support web sockets, but only has TypeScript and JS client support.
But it looks like someone already raised to have Dart support here https://github.com/improbable-eng/grpc-web/issues/46, and its not too hard rom looking at the code

sigurdm:
> @sigurdm Can you explain by what you mean "without dart:html" please ?

I can try to elaborate a little:

When dart is run on a mobile device as flutter or on the command line via the VM it cannot import `dart:html`. (Only dart compiled to JavaScript and run in the browser as in flutter-web has access to this library)

Currently the dart grpc-web implementation relies on HttpRequest (the same as XmlHttpRequest in javascript) from `dart:html`. 

It should be possible to instead use https://pub.dev/packages/http for making the requests. It is supported both on the VM and in the browser.

Non-web grpc is based on http2 that is only available in the vm settings (flutter and command line).

winwisely99:
Any examples of this anywhere @sigurdm 

We are using flutter web, mobile and desktop with GRPC.

sigurdm:
Not sure I understood. Examples of what?

nichite:
One this this breaks is the ability to send credentials with web requests. This was previously possible with `dart:html` import, but I don't believe I see such options anywhere in the `Request` interface from the HTTP library.

songfei:
I have implemented this, and it can already work on my project. I haven't written the test yet. I will submit it to PR after I finish it.

https://github.com/songfei/grpc-dart

lukepighetti:
@mraleph @songfei any chance we can put some heads together to triage the above solution?

MaMrEzO:
Our endpoint just provide `grpc-web` and I cannot make it work in Flutter(Not Flutter-Web), `grpc-web` does not mean its calling from the web or Javascript that's a protocol, do not limit it with xhr!

MaMrEzO:
[...]

PR #557, opened 2022-06-03
Support for http based grpc-web for mobile

otakuto:
https://github.com/grpc/grpc-dart/pull/287
https://github.com/grpc/grpc-dart/pull/351
This will be a re-attempting of the above.

Thanks for your review.

### DIFF
--- a/lib/src/client/transport/http_transport.dart
+++ b/lib/src/client/transport/http_transport.dart
@@ -1,0 +1,183 @@
+// Copyright (c) 2018, the gRPC project authors. Please see the AUTHORS file
+// for details. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'dart:async';
+import 'dart:typed_data';
+
+import 'package:http/http.dart' as http;
+
+import '../../client/call.dart';
+import '../../shared/message.dart';
+import '../../shared/status.dart';
+import '../connection.dart';
+import 'cors.dart' as cors;
+import 'transport.dart';
+import 'web_streams.dart';
+
+const _contentTypeKey = 'Content-Type';
+
+class HTTPTransportStream implements GrpcTransportStream {
+  final ErrorHandler _onError;
+  final Function(HTTPTransportStream stream) _onDone;
+  final StreamController<ByteBuffer> _incomingProcessor = StreamController();
+  final StreamController<GrpcMessage> _incomingMessages = StreamController();
+  final StreamController<List<int>> _outgoingMessages = StreamController();
+
+  @override
+  Stream<GrpcMessage> get incomingMessages => _incomingMessages.stream;
+
+  @override
+  StreamSink<List<int>> get outgoingMessages => _outgoingMessages.sink;
+
+  HTTPTransportStream(client, request,
+      {required ErrorHandler onError, required onDone})
+      : _onError = onError,
+        _onDone = onDone {
+    () async {
+      _outgoingMessages.stream
+          .map(frame)
+          .listen((data) => request.sink.add(data), onDone: () {
+        request.sink.close();
+      }, cancelOnError: true);
+
+      final response = await client.send(request);
+
+      _incomingMessages.add(GrpcMetadata(response.headers));
+
+      response.stream.listen((event) async {
+        if (_incomingProcessor.isClosed) {
+          return;
+        }
+        _incomingProcessor.add(Uint8List.fromList(event).buffer);
+      }, onDone: () {
+        _onRequestDone(response);
+        _close();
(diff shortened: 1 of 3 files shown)


devoncarew:
@sigurdm @mraleph - who would be best to review this?

mraleph:
@otakuto thanks for taking your time to contribute. Could you try to fix EasyCLA check? We can't look at the code before it is fixed.

mraleph:
@otakuto Could you outline a specific use case which makes you interested in using gRPC-Web from a mobile application? Any specific reason that prevents you from using the native gRPC protocol instead?

/cc @wenbozhu

otakuto:
> @otakuto thanks for taking your time to contribute. Could you try to fix EasyCLA check? We can't look at the code before it is fixed.

I fixed it.

otakuto:
> @otakuto Could you outline a specific use case which makes you interested in using gRPC-Web from a mobile application? Any specific reason that prevents you from using the native gRPC protocol instead?
> 
> /cc @wenbozhu

In this example, there is a reason why it is difficult for a microcontroller to support HTTP2 when communicating with a microcontroller such as ESP32 and mobile, for example.
Another advantage is that it is possible to implement a pure dart client for mobile and web respectively.

mraleph:
@otakuto 

> In this example, there is a reason why it is difficult for a microcontroller to support HTTP2 when communicating with a microcontroller such as ESP32 and mobile.

Makes sense. HTTP2 stack might be quite large so it does not necessarily fit on a microcontroller. 

> Another advantage is that it is possible to implement a pure dart client for mobile and web respectively.

Could you elaborate what you mean by this?

One of the problems with merging this (and officially supporting gRPC-Web protocol on native targets) is that it would block us from migrating our gRPC-Web implementation to use official JavaScript implementation underneath, which is something that we have been considering to reduce our maintenance burden.

johynpapin:
Hello !

I want to share another use case for this feature.

As written in #216, another use case is to be able to use the gRPC-Web protocol to bypass the limitations of some cloud platforms, such as DigitalOcean's App Platform service which does not allow to expose a gRPC service.

mreimer:
I'd like to add another vote in support of this PR. AWS lambda doesn't support HTTP2, and while I'm not familiar with other FaaS environments I expect lambda isn't the only one with this constraint. I appreciate the need to control your maintenance surface area but I think it would be a shame to commit to never supporting grpc-web on native targets.

For what it's worth I've locally rebased this PR on 3.1.0 and tested it for my use case, LGTM.

johynpapin:
Hello !

We have tested these changes on our API and everything seems OK on our side too.

Maybe another solution would be to export what is needed to be able to offer this functionality in another package?

lib/src/client/transport/http_transport.dart
```diff
@@ -0,0 +1,183 @@
...
+  final StreamController<GrpcMessage> _incomingMessages = StreamController();
+  final StreamController<List<int>> _outgoingMessages = StreamController();
+
+  @override
+  Stream<GrpcMessage> get incomingMessages => _incomingMessages.stream;
+
+  @override
+  StreamSink<List<int>> get outgoingMessages => _outgoingMessages.sink;
+
+  HTTPTransportStream(client, request,
```
mosuem, line 44: The arguments should be typed.

lib/src/client/transport/http_transport.dart
```diff
@@ -0,0 +1,183 @@
...
+import '../../shared/message.dart';
+import '../../shared/status.dart';
+import '../connection.dart';
+import 'cors.dart' as cors;
+import 'transport.dart';
+import 'web_streams.dart';
+
+const _contentTypeKey = 'Content-Type';
+
+class HTTPTransportStream implements GrpcTransportStream {
```
mosuem, line 31: Rename this and similar to ``Http...` for consistency.

lib/src/client/transport/http_transport.dart
```diff
@@ -0,0 +1,183 @@
...
+    _onDone(this);
+  }
+
+  @override
+  Future<void> terminate() async {
+    _close();
+  }
+}
+
+class HTTPClientConnection extends ClientConnection {
```
mosuem, line 110: Why `extend` and not `implement`?

lib/src/client/transport/http_transport.dart
```diff
@@ -0,0 +1,183 @@
...
+    if (callOptions is WebCallOptions &&
+        callOptions.bypassCorsPreflight == true) {
+      requestUri = cors.moveHttpHeadersToQueryParam(metadata, requestUri);
+    }
+
+    final request = http.StreamedRequest('POST', requestUri);
+
+    for (final header in metadata.keys) {
+      request.headers[header] = metadata[header]!;
+    }
```
mosuem, line 146: Should the other headers also be set as in the `XhrClientConnection`?

lib/src/client/transport/http_transport.dart
```diff
@@ -0,0 +1,183 @@
...
+        }
+        _incomingProcessor.add(Uint8List.fromList(event).buffer);
+      }, onDone: () {
+        _onRequestDone(response);
+        _close();
+      }, onError: (error, stackTrace) {
+        if (_incomingProcessor.isClosed) {
+          return;
+        }
+        _onError(GrpcError.unavailable('HTTPConnection connection-error'),
```
mosuem, line 71: See above, also `Http`

mosuem (changes requested):
Thanks for the contribution! Tests are missing and a few smaller comments.

mosuem:
Closing as this has been waiting for changes for a long time.

Issue #558, opened 2022-06-17
Sending data after the client got closed crashed the program instead of sending an error.


guyluz11:
<short description>
Sending data after the client got closed crashed the program instead of sending an error upwards for the programmer to handle it.

<version of the grpc-dart packages used; see your `pubspec.lock` file>
`grpc: ^3.0.2`

## Repro steps

1. step1
1. step2
1. step3

Expected result: <describe what should have happened>
The error gets passed up and can be captured using handleError or try-catch 

Actual result: <describe what actually happened>
The program crash without a way to capture the error

## Details

<Include any other relevant details, logs, etc.>

> 
> Unhandled exception:
> Bad state: Cannot add event after closing
> #0      _StreamController.add (dart:async/stream_controller.dart:595:24)
> #1      _StreamSinkWrapper.add (dart:async/stream_controller.dart:867:13)
> #2      TransportStream.sendHeaders (package:http2/transport.dart:147:22)
> #3      ServerHandler_.sendTrailers (package:grpc/src/server/handler.dart:361:13)
> #4      ServerHandler_._sendError (package:grpc/src/server/handler.dart:410:5)
> #5      ServerHandler_._onResponse (package:grpc/src/server/handler.dart:287:7)
> #6      _RootZone.runUnaryGuarded (dart:async/zone.dart:1618:10)
> #7      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:341:11)
> #8      _DelayedData.perform (dart:async/stream_impl.dart:591:14)
> #9      _StreamImplEvents.handleNext (dart:async/stream_impl.dart:706:11)
> #10     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:663:7)
> #11     _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
> #12     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
> #13     _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:122:13)
> #14     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:193:5)
> 
> Process finished with exit code 255

PR #559, opened 2022-06-17
Fixing crash when sending headers to closed client (#558)

guyluz11:
Fix for #558


mraleph:
I don't think just swallowing exception is the right fix for this problem. 

You could start by writing a test to demonstrate an issue and then we could try to figure out the correct solution.

Issue #428, opened 2021-01-24
Expose onStateChanged from Http2ClientConnection


long1eu:
Right now this field is marked as `visibleForTesting`. But it would be awesome to monitor the state of the stream and even expose a `Stream<ConnectionState>`.

shekohex:
Yes, please!
one use case is when the connection is restored we can dequeue the failed/pending requests and send them again.

Cobinja:
This would be great to have on the channel, similar to how it's done in the python implementation (https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_channel.py#L1493).

Ludonope:
Any update on that?

It seems pretty essential to be able to know the state of the current connection.
I'm using gRPC for a Flutter app and being able to show the current connection state to the user is an important feature.

micheljung:
Looks like this has been fixed with #565 (3.1.0).

PR #563, opened 2022-06-22
Add onConnectionStateChanged on ClientConnection

Ludonope:
This PR adds a callback on `ClientConnection` to get updates when the `ConnectionState` changes.

Usage:
```dart
final conn = await clientChannel.getConnection();
conn.onConnectionStateChanged = (ConnectionState prevState, ConnectionState state) {
  // ... do whatever.
}
```

This is a first solution for issue #428. It's not perfect yet and could be improved upon, I'm open to critics/suggestions.

## 👀 Why is it important?

A modern app (particularly GUI) should be able to provide feedback to user related to the state of the connection. We don't want the user to wait until they interact to tell them that "oh actually the connection is down", we need to tell them as soon as possible.

## 🛠 Possible improvements

1. Maybe we could set that callback as a parameter of the `ClientChannel` ? That way we wouldn't have to force the creation of the connection just to set it.

2. Is the previous state necessary ? I put it there as I feel like it's important, going from `idle` to `connected` is not the same as going from `transientFailure` to `connected`. But we could also leave that to the user to store the previous state 🤔

### DIFF
--- a/lib/src/client/connection.dart
+++ b/lib/src/client/connection.dart
@@ -57,4 +57,7 @@ abstract class ClientConnection {
   /// All open calls are terminated immediately, and no further calls may be
   /// made on this connection.
   Future<void> terminate();
+
+  void Function(ConnectionState prevState, ConnectionState state)?
+      onConnectionStateChanged;
 }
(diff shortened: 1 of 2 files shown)


Cobinja:
If this is merged then do not merge #565.

mraleph:
Thanks for contribution @Ludonope! 

I think stream based API proposed by @Cobinja is more in line with Dart style (it also allows for multiple listeners out of the box, etc). Let's try to land #565 instead of this one.

PR #565, opened 2022-06-26
Expose onConnectionStateChanged for channels

Cobinja:
This offers a Stream<ConnectionState> on channels. Only for Http2Channel, not for web usage.

Usage:
clientChannel.onConnectionStateChanged.listen(...)

This is an alternative to #563. Only merge one of these.

### DIFF
--- a/lib/src/client/channel.dart
+++ b/lib/src/client/channel.dart
@@ -42,7 +42,7 @@ abstract class ClientChannel {
 
   /// Stream to listen for connection changes
   /// 
-  /// This returns a broadcast stream that can be listened to for conenction changes.
+  /// This returns a broadcast stream that can be listened to for connection changes.
   Stream<ConnectionState> get onConnectionStateChanged;
 }
 
--- a/lib/src/client/channel.dart
+++ b/lib/src/client/channel.dart
@@ -15,8 +15,6 @@
 
 import 'dart:async';
 
-import 'package:meta/meta.dart';
-
 import '../shared/profiler.dart';
 import '../shared/status.dart';
 
@@ -43,7 +41,7 @@ abstract class ClientChannel {
       ClientMethod<Q, R> method, Stream<Q> requests, CallOptions options);
 
   /// Stream of connection state changes
-  /// 
+  ///
   /// This returns a broadcast stream that can be listened to for connection changes.
   /// Note: on web channels, this will not yield any values.
   Stream<ConnectionState> get onConnectionStateChanged;
@@ -55,8 +53,7 @@ abstract class ClientChannelBase implements ClientChannel {
   late ClientConnection _connection;
   var _connected = false;
   bool _isShutdown = false;
-  @protected // no need for this to be public but needed in http2_channel.dart
-  final StreamController<ConnectionState> connectionStateStreamController =
+  final StreamController<ConnectionState> _connectionStateStreamController =
       StreamController.broadcast();
 
   ClientChannelBase();
@@ -67,7 +64,7 @@ abstract class ClientChannelBase implements ClientChannel {
     _isShutdown = true;
     if (_connected) {
       await _connection.shutdown();
-      await connectionStateStreamController.close();
+      await _connectionStateStreamController.close();
     }
   }
 
@@ -76,7 +73,7 @@ abstract class ClientChannelBase implements ClientChannel {
     _isShutdown = true;
     if (_connected) {
       await _connection.terminate();
-      await connectionStateStreamController.close();
+      await _connectionStateStreamController.close();
     }
   }
 
@@ -89,6 +86,12 @@ abstract class ClientChannelBase implements ClientChannel {
     if (_isShutdown) throw GrpcError.unavailable('Channel shutting down.');
     if (!_connected) {
       _connection = createConnection();
+      _connection.onStateChanged = (state) {
+        if (_connectionStateStreamController.isClosed) {
+          return;
+        }
+        _connectionStateStreamController.add(state);
+      };
       _connected = true;
     }
     return _connection;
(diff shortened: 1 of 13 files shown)


lib/src/client/http2_channel.dart
```diff
@@ -31,27 +33,89 @@ class ClientChannel extends ClientChannelBase {
...
+        return;
+      }
+      connectionStateStreamController.add(c.state);
+    };
+    return connection;
   }
+
+  @override
+  Future<void> shutdown() {
+    return super.shutdown()
```
mraleph, line 57: This code should probably use `whenComplete` instead of `then`, otherwise we will not close controller if `shutdown` throws an error. 

Alternatively you could write (which I personally consider easier to read): 

```dart
Future<void> shutdown() async {
  try {
    return super.shutdown();
  } finally {
    connectionStateStreamController.close();
  }
}
```

lib/src/client/http2_channel.dart
```diff
@@ -31,27 +33,89 @@ class ClientChannel extends ClientChannelBase {
...
+    .then((value) {
+      connectionStateStreamController.close();
+      return value;
+    });
+  }
+
+  @override
+  Future<void> terminate() {
```
mraleph, line 65: Same comment as above.

lib/src/client/channel.dart
```diff
@@ -39,6 +39,11 @@ abstract class ClientChannel {
   /// Initiates a new RPC on this connection.
   ClientCall<Q, R> createCall<Q, R>(
       ClientMethod<Q, R> method, Stream<Q> requests, CallOptions options);
+
+  /// Stream to listen for connection changes
```
mraleph, line 43: Maybe: `Stream of connection state changes`.

lib/src/client/http2_channel.dart
```diff
@@ -31,27 +33,89 @@ class ClientChannel extends ClientChannelBase {
   final Object host;
   final int port;
   final ChannelOptions options;
+  final StreamController<ConnectionState> connectionStateStreamController =
```
mraleph, line 36: Can this live in `ClientChannelBase` instead? Then you don't need any of the duplicated code below. 

You would need to add a member to `ClientChannel` though to expose state changes in some form.

lib/src/client/http2_channel.dart
Cobinja: Done in the new commit

lib/src/client/channel.dart
Cobinja: Done in the new commit

lib/src/client/http2_channel.dart
Cobinja: Done in the new commit

lib/src/client/http2_channel.dart
Cobinja: Done in the new commit

lib/src/client/http2_channel.dart
mraleph: My suggestion was to move all of the code to the base class (this includes code from `terminate` and `shutdown` as well as attaching status change callback to the connection object).

mraleph: No rush.

lib/src/client/channel.dart
```diff
@@ -39,6 +41,12 @@ abstract class ClientChannel {
   /// Initiates a new RPC on this connection.
   ClientCall<Q, R> createCall<Q, R>(
       ClientMethod<Q, R> method, Stream<Q> requests, CallOptions options);
+
+  /// Stream of connection state changes
+  /// 
+  /// This returns a broadcast stream that can be listened to for connection changes.
+  /// Note: on web channels, this will not yield any values.
```
mraleph, line 48: Theoretically we could provide implementation for the web channels as well (as underlying XHR transport has `onReadyStateChange`). One source of complexity here would be that on Web one `ClientChannel` corresponds to many connections. You don't need to fix it now but we should write code in a generic fashion (e.g. have `ClientConnection.onStateChange` as suggested) so that we can easily extend it later if needed.

lib/src/client/http2_connection.dart
```diff
@@ -43,7 +43,6 @@ class Http2ClientConnection implements connection.ClientConnection {
 
   connection.ConnectionState _state = ConnectionState.idle;
 
-  @visibleForTesting
   void Function(Http2ClientConnection connection)? onStateChanged;
```
mraleph, line 46: I feel like you could still add `onStateChanged` to `ClientConnection`:

```dart
class ClientConnection {
  /// Set state change listener for this connection. The given callback will be 
  /// invoked when the state of this connection changes.
  set onStateChanged(void Function(ConnectionState) cb);
}
```

(Note: I changed signature because this will make code simpler.).

This setter can be (for now) an no-op for XHR based connection. 

This would allow you to avoid any platform specific and duplicated code in `lib/src/client/http2_channel.dart`.

mraleph (changes requested):
Please change version number in `pubspec.yaml` to `3.1.0-dev` and add corresponding `CHANGELOG.md` entry describing the change.

mraleph:
Thanks @Cobinja. I think this is good to go. I have refactored things a bit more (you introduced the field I asked for - but did not actually make use of it to share repeating code). You can take a look at the final version.

mraleph (approved):

Cobinja:
My thought process was "Web channels might need a completely different implementation, so just provide the common API". I'm fine with your changes.

andreadaoud:
Hi,

Just a ping on this PR. What's blocking this from being merged?

natebosch:
This was a breaking change - there are internal implementations of `ClientConnection` which are broken because they no longer implement the entire interface.

Issue #573, opened 2022-09-02
Deadline exceeded errors do not contain stacktrace 


tarobins:
No stack trace shown for dealine exceeded errors.

(version: "3.1.0-dev")

## Repro steps

1. checkout timeout branch from https://github.com/tarobins/grpc-dart/tree/timeout 
1. start helloworld example server
1. run helloworld example client

Expected result: 

```
Caught error: gRPC Error (code: 4, codeName: DEADLINE_EXCEEDED, message: Deadline exceeded, details: null, rawResponse: null, trailers: {})
Stack trace

That's all folks.
robinsontom-macbookpro1 example/helloworld$ dart bin/client.dart
Caught error: gRPC Error (code: 4, codeName: DEADLINE_EXCEEDED, message: Deadline exceeded, details: null, rawResponse: null, trailers: {})
Stack trace
#0      ClientCall._onTimedOut (package:grpc/src/client/call.dart:303:43)
#1      Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18:15)
#2      _Timer._runTimers (dart:isolate-patch/timer_impl.dart:398:19)
#3      _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:429:5)
#4      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)

That's all folks.
```

Actual result: 

```
Caught error: gRPC Error (code: 4, codeName: DEADLINE_EXCEEDED, message: Deadline exceeded, details: null, rawResponse: null, trailers: {})
Stack trace

That's all folks.
```

PR #574, opened 2022-09-02
Add stacktrace to deadline exceeded error.

tarobins:
BUG: https://github.com/grpc/grpc-dart/issues/573

### DIFF
--- a/lib/src/client/call.dart
+++ b/lib/src/client/call.dart
@@ -300,7 +300,7 @@ class ClientCall<Q, R> implements Response {
   void _onTimedOut() {
     final error = GrpcError.deadlineExceeded('Deadline exceeded');
     _finishTimelineWithError(error, _requestTimeline);
-    _responses.addError(error);
+    _responses.addError(error, StackTrace.current);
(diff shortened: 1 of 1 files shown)


tarobins:
If you have suggestions on how to test this change, please let me know.

mraleph:
@tarobins does attaching this stack trace actually communicate any additional information? It's always going to be the same: a bunch of internal frames related to the timer implementation and the `ClientCall._onTimedOut` frame at the top.

tarobins:
@mraleph 

Yes, you are correct.  I thought I saw something useful when I was testing the change, but perhaps I was just overly optimistic.  

On the other hand, I spent a bunch of time trying to figure out why there was no stack trace.  Had I seen the, even not particularly useful, one that this change provides, I would have at least understood where the error was appearing in the grpc library code rather than chasing why there was no stack trace.  That said, I'm going to see if I can improve the situation more, so no need for you to review this change further right now.  

Thanks.

mraleph:
Closing as stale for now.

Issue #576, opened 2022-09-08
HTTP/2 error: Stream error: Stream was terminated by peer (errorCode: 1) when using Unix sockets on OSX


denizt:
All gRPC calls fail when using UNIX sockets on mullvad-daemon

grpc: 3.0.2

## Repro steps

1. Get the example project attached
1. pub get
1. Execute unix_client

Expected result: TunnelState indicating the state of the VPN

Actual result: HTTP/2 error: Stream error: Stream was terminated by peer (errorCode: 1)

## Details
Might be related to https://github.com/grpc/grpc-dart/issues/572
The server is built with https://github.com/hyperium/tonic and works as expected for both rust and grpc-python

[mullvad_daemon_test.zip](https://github.com/grpc/grpc-dart/files/9523908/mullvad_daemon_test.zip)

faern:
The server also works as expected with a `gprc-js` client, as that's what the GUI frontend is using. https://github.com/mullvad/mullvadvpn-app/blob/dca39b225e54eecc2309bff371e7f6c5262f5203/gui/src/main/daemon-rpc.ts#L1

mraleph:
> [mullvad_daemon_test.zip](https://github.com/grpc/grpc-dart/files/9523908/mullvad_daemon_test.zip)

Could you please put your reproduction into some github/gitlab repo and license it under some open-source license of your choice? We can't look at any code that does not have a clear license unfortunately.

denizt:
Sure, added it to the examples in the fork https://github.com/denizt/grpc-dart/tree/mullvad-testcase/example

Same license as the repo, Apache-2.0

mraleph:
I took a quick look. Enabling verbose logging on the server side reveals that it (rightfully so) does not like authority field which `grpc-dart` sends. You can work-around by doing `ChannelCredentials.insecure(authority: 'localhost')`. 

I will make PR which would make `grpc-dart` default to `localhost` authority when connecting over UDS.

PR #577, opened 2022-09-08
Change default authority for UDS connections.

mraleph:
We were using path to the UDS socket itself, which is incorrect authority value and was tripping checks in some HTTP2 protocol implementations.

Instead default authority to localhost, which seems like a sensible value.

Fixes https://github.com/grpc/grpc-dart/issues/576


osa1 (approved):
LGTM.

PR #581, opened 2022-10-08
Fix settings-frame waiting hack

t123yh:
In https://github.com/dart-lang/http2/commit/9814696a101067c9db9c1b3cc071bfb2c3e44748 , @mraleph Added onInitialPeerSettingsReceived, making it no longer necessary to rely on time hack to wait for settings frame.


mraleph:
You need to delete `_estimatedRoundTripTime`. I tried updating PR myself - but I can't push more changed into it because PR is created from `master` branch which is protected.

PR #520, opened 2021-08-30
Support responseErrorHandler

lejard-h:
Add ability to register a Function call onResponseError, so we can log errors (or anything else)

Fix https://github.com/grpc/grpc-dart/issues/51

Usage
```dart
Server(
  [ service ],
  [],
  codec,
  (error, trace) {
     ...
  }
);
```

PR #588, opened 2022-11-25
Custom Error Handling for Server

### LINKED ISSUE #51: Add logger for debugging
It would be really great if this package would use a (logging-compliant) logger that could be set to get better reporting when things go wrong.

bgetsug:
This picks up where #520 left off and resolves #51.

My particular use case is needing to log (with stack traces) gRPC errors as they occur on the server, such as those having to do with deserialization.

cc: @mraleph

### DIFF
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -35,7 +35,7 @@ class ServerHandlerImpl extends ServiceCall {
   final Service? Function(String service) _serviceLookup;
   final List<Interceptor> _interceptors;
   final CodecRegistry? _codecRegistry;
-  final Function? _responseErrorHandler;
+  final Function? _errorHandler;
 
   // ignore: cancel_subscriptions
   StreamSubscription<GrpcMessage>? _incomingSubscription;
@@ -68,7 +68,7 @@ class ServerHandlerImpl extends ServiceCall {
     this._interceptors,
     this._codecRegistry, [
     this._clientCertificate,
-    this._responseErrorHandler,
+    this._errorHandler,
   ]);
 
   @override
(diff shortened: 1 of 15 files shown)


lib/src/server/handler.dart
```diff
@@ -61,9 +62,14 @@ class ServerHandlerImpl extends ServiceCall {
   Timer? _timeoutTimer;
   final X509Certificate? _clientCertificate;
 
-  ServerHandlerImpl(this._serviceLookup, this._stream, this._interceptors,
-      this._codecRegistry,
-      [this._clientCertificate]);
+  ServerHandlerImpl(
```
mraleph, line 65: Could you convert this one to use named parameter as well, so that call-sites are more readable?

bgetsug: Done 😄

lib/src/server/server.dart
```diff
@@ -87,6 +87,7 @@ class ConnectionServer {
   final Map<String, Service> _services = {};
   final List<Interceptor> _interceptors;
   final CodecRegistry? _codecRegistry;
+  final Function? _errorHandler;
```
mraleph, line 90: This should be tighter typed (here and other places which accept it), e.g. 

```
typedef ErrorHandler = void Function(GrpcError error, StackTrace? trace);
```
otherwise users don't actually know what kind of function they are expected to pass in.

lib/src/server/handler.dart
```diff
@@ -410,7 +423,11 @@ class ServerHandlerImpl extends ServiceCall {
       ..onDone(_onDone);
   }
 
-  void _sendError(GrpcError error) {
+  void _sendError(GrpcError error, [StackTrace? trace]) {
+    if (_errorHandler != null) {
+      _errorHandler!(error, trace);
```
mraleph, line 428: This could simply be: `_errorHandler?.call(error, trace)` - then `if` is not needed.

mraleph (changes requested):
Two final comments - and then I will merge. Thanks a bunch for pushing this through.

lib/src/server/handler.dart
bgetsug: nice...I never think to use `call()`

lib/src/server/server.dart
bgetsug: ✅

long1eu:
Can we have this on pub.dev?

Pingear:
Hello !
where can I find documentation on the new logging functionality?

PR #590, opened 2022-12-01
Expose client IP address to server

bgetsug:
I probably should have opened an issue first, but I wanted to be able to log the client IP address in an `Interceptor`. Unless I missed this already being available somewhere, here's my solution. Thoughts?

### DIFF
--- a/lib/src/server/call.dart
+++ b/lib/src/server/call.dart
@@ -45,7 +45,7 @@ abstract class ServiceCall {
   X509Certificate? get clientCertificate;
 
   /// Returns the IP address of the client, if available
-  InternetAddress? get clientIpAddress;
+  InternetAddress? get remoteAddress;
 
   /// Send response headers. This is done automatically before sending the first
   /// response message, but can be done manually before the first response is
--- a/lib/src/server/call.dart
+++ b/lib/src/server/call.dart
@@ -13,7 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import '../shared/io_bits/io_bits.dart' show X509Certificate;
+import '../shared/io_bits/io_bits.dart' show InternetAddress, X509Certificate;
 
 /// Server-side context for a gRPC call.
 ///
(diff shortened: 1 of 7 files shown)


lib/src/server/call.dart
```diff
@@ -44,6 +44,9 @@ abstract class ServiceCall {
   /// Returns the client certificate if it is requested and available
   X509Certificate? get clientCertificate;
 
+  /// Returns the IP address of the client, if available
+  InternetAddress? get clientIpAddress;
```
mraleph, line 48: Could you rename this into `remoteAddress` to match `dart:io` `Socket` API naming?

mraleph (changes requested):
LGTM with a minor nit. 

(Sorry for the delay - I was away on vacation).

lib/src/server/call.dart
bgetsug: ✅ 😄

mraleph:
LGTM! Thanks.

lib/src/server/server.dart
```diff
@@ -107,16 +107,21 @@ class ConnectionServer {
 
   Service? lookupService(String service) => _services[service];
 
-  Future<void> serveConnection(
-    ServerTransportConnection connection, [
+  Future<void> serveConnection({
```
natebosch, line 110: This was a breaking change.

lib/src/server/server.dart
```diff
@@ -228,13 +236,23 @@ class Server extends ConnectionServer {
...
+
+      final connection = ServerTransportConnection.viaSocket(
+        socket,
+        settings: http2ServerSettings,
+      );
+
+      serveConnection(
+        connection: connection,
+        clientCertificate: clientCertificate,
+        remoteAddress: socket.remoteAddress,
```
whesse, line 254: This line, fetching socket.remoteAddress, has created an uncaught exception in the case where the socket is already closed when we reach this point.

Previously, the server worked if a server connection was requested for a closed socket, and did not throw an exception.  I had a server running for 16 months without failing using the previous code, and running it with this code failed in less than 8 hours.

PR #620, opened 2023-04-19
Add a ChannelShutdownHandler callback

natebosch:
Add a way to react to the channel shutdown or termination with an optional callback.

This is not necessarily the API which I would recommend, but this is the API that was already inappropriately shipped and in use internally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 3.1.1-dev
+## 3.2.0-dev
 
 * `ChannelOptions` now exposes `connectTimeout`, which is used on the 
   socket connect. This is used to specify the maximum allowed time to wait
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 * Require Dart 2.17 or greater.
 * Fix issue [#51](https://github.com/grpc/grpc-dart/issues/51), add support for custom error handling.
 * Expose client IP address to server 
+* Add a `channelShutdownHandler` argument to `ClientChannel` and the subclasses.
+  This callback can be used to react to channel shutdown or termination.
 
 ## 3.1.0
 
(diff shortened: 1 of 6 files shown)


natebosch:
@kevmoo @mraleph

lib/src/client/channel.dart
```diff
@@ -22,6 +22,9 @@ import 'call.dart';
 import 'connection.dart';
 import 'method.dart';
 
+/// Handler to be called when the channel is shut down.
+typedef ChannelShutdownHandler = void Function();
```
kevmoo, line 26: Do we need such a simple `typedef`?

natebosch: No, it's a bad API.

It's also breaking for google3 to remove it.

natebosch: I can give a shot at cleanup and see how long it takes to drop it.

natebosch: It looks relatively easy to remove internally, I'll send out a CL and hopefully the owners of the code with usage will quickly approve.

If not, I may also be able to make the typedef internal only with a patch file.

natebosch: I was able to remove the typedef internally.

kevmoo (approved):

natebosch:
@kevmoo  - can you merge this?

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 * Require Dart 2.17 or greater.
 * Fix issue [#51](https://github.com/grpc/grpc-dart/issues/51), add support for custom error handling.
 * Expose client IP address to server 
+* Add a `channelShutdownHandler` argument to `ClientChannel` and the subclasses.
```
kevmoo, line 11: Ugh. Does this make it 3.2.0-dev?

natebosch: Yes, good catch.

PR #633, opened 2023-06-06
Switch to Dart 3

### LINKED ISSUE #613: Dart 3 compatibility: `_ResponseMixin` should be changed to be an actual `mixin`
This definition should be changed to `mixin`:
https://github.com/grpc/grpc-dart/blob/eb833d3f1c4ee9b588fe6bf5e18ed476edb938f9/lib/src/client/common.dart#L80

Otherwise this will fail to build if Dart language version 3 is used.
https://github.com/grpc/grpc-dart/blob/eb833d3f1c4ee9b588fe6bf5e18ed476edb938f9/lib/src/client/common.dart#L46

The change will be fully backwards-compatible (it's a private symbol), so it can be applied sooner rather than later.

mosuem:
Fixes #613 .

### DIFF
--- a/test/timeline_test.dart
+++ b/test/timeline_test.dart
@@ -148,6 +148,6 @@ void main([args = const <String>[]]) {
     checkReceiveEvent(events);
     checkReceiveMetaDataEvent(events);
     checkFinishEvent(events);
-    vmService.dispose();
+    await vmService.dispose();
   });
 }
--- a/test/timeline_test.dart
+++ b/test/timeline_test.dart
@@ -87,13 +87,9 @@ void checkStartEvent(List<Map> events) {
   expect(e.length, 2);
 
   expect(e[0]['name'], 'gRPC Request: $path');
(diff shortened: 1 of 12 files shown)


test/timeline_test.dart
```diff
@@ -64,78 +67,13 @@ class FixedConnectionClientChannel extends ClientChannelBase {
...
-Future<void> testee() async {
+Future<VmService> testee() async {
+  isTimelineLoggingEnabled = true;
+  final info = await dev.Service.getInfo();
+  final uri = info.serverWebSocketUri!;
+  final vmService = await vmServiceConnectUri(uri.toString());
   final server = Server.create(services: [TestService()]);
   await server.serve(address: 'localhost', port: 0);
-  isTimelineLoggingEnabled = true;
-  timelineTaskFactory = fakeTimelineTaskFactory;
```
mraleph, line 138: You should probably remove this field and all its uses as well.

mraleph (approved):

PR #634, opened 2023-06-09
Keepalive

### LINKED ISSUE #157: Add KeepAlive support
I have a gRPC service (in dart) that provides a rpc method returning a stream of messages.

In some cases, this stream may stay without message during a long time. Without activity on the connection during 2 minutes a Firewall kills the connection and the client has to reconnect.

So it would be great to provide a way to send hearbeat.

This problem seems to be addressed by https://github.com/grpc/grpc/blob/master/doc/keepalive.md

mosuem:
Fixes #157.

For now, only handles pings from the client to the server.

Relies on https://github.com/dart-lang/http2/pull/127, as http2 needs to expose when a ping is received, instead of just silently acknowledging it.


lib/src/shared/keepalive.dart
```diff
@@ -0,0 +1,254 @@
...
+  /// Transport goes idle after ping has been sent.
+  idleAndPingSent,
+
+  /// The transport has been disconnected. We won't do keepalives any more.
+  disconnected;
+}
+
+/// A keep alive "manager", deciding when to send pings or shutdown based on the
+/// [ClientKeepAliveOptions].
+class ClientKeepAlive {
```
mraleph, line 69: This class can benefit from some form of state transition diagram - otherwise it is very hard to follow how states transition into each other. 

Additionally it is rather confusing which of the `pingTimer` and `pingTimer` can be active at any given state moment. I wonder if there is a better way to express it, e.g. by gluing timers and states together. 

Maybe (just a sketch, maybe some other way to express the same is possible):

```dart
sealed class ClientKeepAliveState {
  void detach();
}

final class StatePingSent extends ClientKeepAliveState {
  final Timer shutdown;
  bool idle;
  
  void detach() {
    shutdown.cancel();
  }
  
  ClientKeepAliveState handle(Event event) => switch (event) {
    Event.frame when !idle => StatePingScheduled(...),
    Event.frame when idle => StateIdle(shutdown: Timer(...)), // inherit the shutdown
  };
} 

final class StatePingScheduled extends ClientKeepAliveState {
  final Timer ping;
} 

class ClientKeepAlive {
  void _handle(Event e) {
    final newState = _state.handle(e);
    if (!identical(newState, _state)) {
      _state.detach();
    }
    _state = newState;
  }

  void onFrameReceived() {
    _handle(Event.frame);  
  }
}
```

Maybe this will end up extremely verbose (probably), but at least it would make a lot of transitions very easy to reason about.

mraleph (changes requested):
I think this is looking okay, though some of the state transition logic is a bit hard to review. 

Are there interop tests for this functionality?

lib/src/shared/keepalive.dart
mosuem: I tried something of the sort before, but I found it even more confusing due to the large amount of code needed. But I will try if I can find another solution.

mosuem:
There are no interop tests for this AFAIK.

lib/src/server/server.dart
```diff
@@ -166,7 +183,8 @@ class Server extends ConnectionServer {
   /// Create a server for the given [services].
   @Deprecated('use Server.create() instead')
   Server(
-    super.services, [
+    super.services,
+    super.keepAlive, [
```
mraleph, line 187: This is breaking change. Consider just not adding support for keepAlive here and intialize it with default value instead.

lib/src/client/client_keepalive.dart
```diff
@@ -0,0 +1,241 @@
...
+  Duration get _pingInterval => _options.pingInterval ?? Duration(days: 365);
+  final Map<Type, KeepAliveState? Function(KeepAliveState, _KeepAliveEvent)>
+      _stateTransform = {};
+
+  ClientKeepAlive({
+    required ClientKeepAliveOptions options,
+    required this.ping,
+    required this.onPingTimeout,
+  }) : _options = options {
+    _addEventHandler<Idle>((state, event) {
```
mraleph, line 123: Any specific reason that these is not just a method on the state itself. Using `runtimeType` + map for state machine feels a bit too dynamic.

mraleph (approved):
This looks great! Just a minor comment about state machine.

lib/src/client/client_keepalive.dart
mosuem: I didn't know how to wire through needed properties such as `ping` or `options` from `ClientKeepAlive` without making the code ugly or complicated. I thought this was a nice alternative.

mosuem: I am open for other ideas though :)

Issue #617, opened 2023-04-08
Server.create constructor is deprecated, need to update sample code


craiglabenz:
Existing sample code suggests creating servers like so:

```dart
grpc.Server.create(services: [RouteGuideService()])
```

But that seems like it should be updated to:

```dart
grpc.Server([RouteGuideService()])
```

mraleph:
Hmm, I am puzzled. It is the other way around and samples were in fact updated to use `Server.create`.

```
  /// Create a server for the given [services].
  @Deprecated('use Server.create() instead')
  Server(...);
```

We do need to update the basics guide though which is hosted in some other repo: https://grpc.io/docs/languages/dart/basics/

broderix:
UP.
@mraleph , thanks for response. I think you should bump new version 3.1.1 from master. Because example correct with version in master, that uses  `Server.create()` .

mraleph:
@mosuem could you take a look at releasing the new version? (maybe we can even use this as an opportunity to configure publishing from GitHub actions for this repo). 

(I have created PR to update docs: https://github.com/grpc/grpc.io/pull/1145)

PR #637, opened 2023-06-13
Prepare release of `3.2.0`

mosuem:
Part of the fix for #617, the other being https://github.com/grpc/grpc.io/pull/1145.


lib/src/server/server.dart
```diff
@@ -163,15 +163,6 @@ class Server extends ConnectionServer {
   ServerSocket? _insecureServer;
   SecureServerSocket? _secureServer;
 
-  /// Create a server for the given [services].
```
mraleph, line 166: This probably requires bump to `4.0.0` and the same below.

mraleph (approved):
LGTM though I think we should not remove APIs unless we are incrementing to 4.0.0

lib/src/server/server.dart
mosuem: Right..

PR #641, opened 2023-06-17
support latest pkg:protobuf

kevmoo:


lib/src/client/common.dart
```diff
@@ -66,7 +66,8 @@ class ResponseFuture<R> extends DelegatingFuture<R>
 }
 
 /// A gRPC response producing a stream of values.
-class ResponseStream<R> extends StreamView<R> with _ResponseMixin<dynamic, R> {
+base class ResponseStream<R> extends StreamView<R>
```
kevmoo, line 69: Got yelled at about this...

domesticmouse (approved):

.github/workflows/dart.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        sdk: [stable, dev]
+        sdk: [3.0.0, dev]
```
domesticmouse, line 19: It may be worth reintroducing `stable` when it becomes 3.1.

domesticmouse (approved):

.github/workflows/dart.yml
domesticmouse: Arguably `stable` is already useful at 3.0.5.

PR #642, opened 2023-06-19
Make ResponseStream a regular class

aran:
Reverts part of https://github.com/grpc/grpc-dart/commit/d9553ca73f66116f7ad14fff5d0e4814253311a4

Making `ResponseStream` a `base` class is not backward-compatible and makes it hard or impossible to work around https://github.com/grpc/grpc-dart/issues/413. 

Arguably, ResponseStream and ResponseFuture are serving two purposes: "The interface interceptors need to work with" and the base of implementations that have `ClientCall` state. A backwards-compatible future change could in principle separate an `interface` and a `base` class to decouple these.

### DIFF
--- a/lib/src/client/common.dart
+++ b/lib/src/client/common.dart
@@ -66,8 +66,7 @@ class ResponseFuture<R> extends DelegatingFuture<R>
 }
 
 /// A gRPC response producing a stream of values.
-base class ResponseStream<R> extends StreamView<R>
-    with _ResponseMixin<dynamic, R> {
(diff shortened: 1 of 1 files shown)


kevmoo:
The problem: this now fails with Dart 3.1-dev – ugh.

kevmoo:
This *will* break with the latest dev release, but we can cross that bridge soon.

kevmoo (approved):

kevmoo:
This will NOT break. I have an old SDK setup locally. User error. Sorry!

PR #657, opened 2023-08-04
Add proxy functionality

### LINKED ISSUE #33: Add support for proxies
Since it's super annoying to debug network protocols without running them through [Charles](https://www.charlesproxy.com/).

Should respect `http_proxy` environment variable (and friends).

mosuem:
Closes #33.

### DIFF
--- a/lib/src/client/http2_connection.dart
+++ b/lib/src/client/http2_connection.dart
@@ -360,9 +360,9 @@ class SocketTransportConnector implements ClientTransportConnector {
   final ChannelOptions _options;
   late Socket socket;
 
-  Proxy get proxy => _options.proxy;
-  Object get host => proxy.isDirect ? _host : proxy.host;
-  int get port => proxy.isDirect ? _port : proxy.port;
+  Proxy? get proxy => _options.proxy;
+  Object get host => proxy == null ? _host : proxy!.host;
+  int get port => proxy == null ? _port : proxy!.port;
 
   SocketTransportConnector(this._host, this._port, this._options)
       : assert(_host is InternetAddress || _host is String);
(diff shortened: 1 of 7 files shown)


lib/src/client/http2_connection.dart
```diff
@@ -351,39 +353,67 @@ class Http2ClientConnection implements connection.ClientConnection {
...
       : assert(_host is InternetAddress || _host is String);
 
   @override
   Future<ClientTransportConnection> connect() async {
     final securityContext = _options.credentials.securityContext;
-    _socket =
-        await Socket.connect(_host, _port, timeout: _options.connectTimeout);
+    socket = await initSocket(host, port);
+
+    Stream<List<int>> incoming;
```
mraleph, line 375: It seems that in both cases code just user `initSocket` + `connectToProxy`, so maybe just do this:

```dart
Future<Stream<int>> connectImpl() async {
  socket = await Socket.connect(connectionHost, connectionPort, timeout: _options.connectTimeout);
  if (proxy.isDirect) {
    return socket;
  }
  
  // We need to ask the proxy to connect to our destination host.
  return connectThroughProxy(); 
}
```

lib/src/client/http2_connection.dart
```diff
@@ -426,6 +456,51 @@ class _SocketTransportConnector implements ClientTransportConnector {
...
+    _sendConnect(headers);
+    await completer.future;
+    return intermediate.stream;
+  }
+
+  void waitFor(Uint8List event, Completer<void> completer) {
+    final response = ascii.decode(event);
+    final lines = response.split('\r\n');
+    // status line
+    final statusLine = lines.first;
```
mraleph, line 496: What happens with the rest of data in this chunk? Is it somehow guaranteed that this chunk will only contain a single line? Needs a comment.

lib/src/client/proxy.dart
```diff
@@ -0,0 +1,23 @@
+class Proxy {
```
mraleph, line 1: I would just use `Proxy?` and `null` to denote direct connection instead of doing `Proxy.direct` and with a bunch of `null` / magical value fields. To avoid `!` everywhere you could do (for example):

```dart
Future<Stream<int>> connectImpl() async {
  socket = await Socket.connect(connectionHost, connectionPort, timeout: _options.connectTimeout);
  final proxy = _proxy;
  if (proxy == null) {
    return socket;
  }
  
  // We need to ask the proxy to connect to our destination host.
  return connectThroughProxy(proxy); // Have connectThroughProxy take an instance of proxy.
}
```

mraleph (approved):
LGTM with some comments. 

I wonder if the functionality for creating HTTP/2 connections over proxies actually belongs to http2 package itself? 

But maybe it's okay to have it here as well for now.

PR #661, opened 2023-08-22
Remove canceled streams

mosuem:
To avoid filling up the list with canceled streams.

### DIFF
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -60,7 +60,6 @@ class ServerHandler extends ServiceCall {
   Map<String, String>? _customTrailers = {};
 
   DateTime? _deadline;
-  bool _isCanceled = false;
   bool _isTimedOut = false;
   Timer? _timeoutTimer;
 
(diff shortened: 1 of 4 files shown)


lib/src/server/handler.dart
```diff
@@ -70,6 +70,10 @@ class ServerHandler extends ServiceCall {
   /// Emits a ping everytime data is received
   final Sink<void>? onDataReceived;
 
+  final StreamController<void> _onCanceledController = StreamController();
```
mraleph, line 73: Consider instead:

```
final Completer<void> _isCanceledCompleter = Completer<void>();

Future<void> onCanceled => _isCanceledCompleter.future;

bool get isCanceled => _isCanceledCompleter.isCompleted;
set isCanceled(bool value) {
  if (!isCanceled) {
    _isCanceledCompleter.complete();
  }
}
```

This way you can get read of the boolean. It also makes more sense to have a `Future` vs `Stream` here because it is a single shot event.

mraleph (approved):
LGTM with a suggestion

Are these code paths covered by the existing tests? I can imagine it is a bit hard to actually check for leaks, but at least I want to make sure that cancelation is exercised.

PR #662, opened 2023-08-23
Fix imports in `http2_connection.dart`

mosuem:
Fix imports and export `Proxy`.

### DIFF
--- a/lib/grpc.dart
+++ b/lib/grpc.dart
@@ -36,6 +36,7 @@ export 'src/client/options.dart'
         BackoffStrategy,
         defaultBackoffStrategy,
         ChannelOptions;
+export 'src/client/proxy.dart' show Proxy;
 export 'src/client/transport/http2_credentials.dart'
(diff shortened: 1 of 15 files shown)


lib/grpc.dart
```diff
@@ -36,6 +36,7 @@ export 'src/client/options.dart'
         BackoffStrategy,
         defaultBackoffStrategy,
         ChannelOptions;
+export 'src/client/proxy.dart' show Proxy;
```
mraleph, line 39: nit: If `Proxy` is a public class then it should probably have some doc comments. 

Also I just noticed that the file is missing copyright header. (Maybe double check that all newly created files have copyright headers).

mraleph (approved):

lib/grpc.dart
mosuem: Time to enable https://github.com/dart-lang/ecosystem/blob/main/.github/workflows/health.yaml in this repo...

PR #664, opened 2023-09-06
Fix regression on fetching the remote address of a closed socket.

mosuem:
Fix of https://github.com/grpc/grpc-dart/pull/590#discussion_r1317152175.

cc @whesse

### DIFF
--- a/lib/src/server/server.dart
+++ b/lib/src/server/server.dart
@@ -244,7 +244,7 @@ class Server extends ConnectionServer {
     Stream<Socket> server;
     final securityContext = security?.securityContext;
     if (securityContext != null) {
-      final __secureServer = await SecureServerSocket.bind(
+      final _server = await SecureServerSocket.bind(
         address ?? InternetAddress.anyIPv4,
         port ?? 443,
         securityContext,
@@ -254,18 +254,18 @@ class Server extends ConnectionServer {
         requestClientCertificate: requestClientCertificate,
         requireClientCertificate: requireClientCertificate,
       );
-      _secureServer = __secureServer;
-      server = __secureServer;
+      _secureServer = _server;
+      server = _server;
     } else {
-      final __insecureServer = await ServerSocket.bind(
+      final _server = await ServerSocket.bind(
         address ?? InternetAddress.anyIPv4,
         port ?? 80,
         backlog: backlog,
         shared: shared,
         v6Only: v6Only,
       );
-      _insecureServer = __insecureServer;
-      server = __insecureServer;
+      _insecureServer = _server;
+      server = _server;
     }
     server.listen((socket) {
       // Don't wait for io buffers to fill up before sending requests.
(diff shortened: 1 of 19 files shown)


lib/src/server/server.dart
```diff
@@ -241,28 +241,33 @@ class Server extends ConnectionServer {
...
     if (securityContext != null) {
-      _secureServer = await SecureServerSocket.bind(
-          address ?? InternetAddress.anyIPv4, port ?? 443, securityContext,
-          backlog: backlog,
-          shared: shared,
-          v6Only: v6Only,
-          requestClientCertificate: requestClientCertificate,
-          requireClientCertificate: requireClientCertificate);
-      server = _secureServer;
+      final __secureServer = await SecureServerSocket.bind(
```
mraleph, line 247: I would avoid `__` if possible. Just give a local variable some normal name without `__` or `_`.

lib/src/server/server.dart
```diff
@@ -320,14 +334,11 @@ class Server extends ConnectionServer {
...
-    final done = _connections.map((connection) => connection.finish()).toList();
-    if (_insecureServer != null) {
-      done.add(_insecureServer!.close());
-    }
-    if (_secureServer != null) {
-      done.add(_secureServer!.close());
-    }
-    await Future.wait(done);
+    await Future.wait([
+      ..._connections.map((connection) => connection.finish()),
```
mraleph, line 338: Consider instead of `...` and `map` just:

```
await Future.wait([
  for (var connection in _connections) connection.finish(),
  // ...
]);
```

lib/src/server/server.dart
```diff
@@ -279,10 +284,19 @@ class Server extends ConnectionServer {
         settings: http2ServerSettings,
       );
 
+      InternetAddress? remoteAddress;
```
mraleph, line 287: Maybe encapsulate this into a helper:

```dart
extension on Socket {
  /// Tries to get a remote address if the socket is still open otherwise returns `null`.
  InternetAddress? remoteAddressOrNull {
    // ...
  }
}
```

mraleph (approved):
LGTM with nits

lib/src/server/server.dart
mosuem: Nice, I didn't know that worked.

mraleph:
Still LGTM

mraleph (approved):

lib/src/server/server.dart
```diff
@@ -320,15 +325,24 @@ class Server extends ConnectionServer {
...
   }
 }
+
+extension on Socket {
+  InternetAddress? get remoteAddressOrNull {
+    try {
+      // Using a try-catch control flow as dart:io Sockets don't expose their
+      // connectivity state.
+      return remoteAddress;
+    } on Exception catch (_) {
```
whesse, line 344: This is good. My fix had SocketException, which is what remoteAddress is documented to throw, but it actually throws an OSError.  Despite its name, OSError is an Exception, not an Error, so this code is right.

PR #666, opened 2023-09-06
Updating handler.dart

T0bst3r:
Adding stack trace to more calls of _sendError by using records.

This should allow for more accurate logging and debugging of issues in custom errorHandlers.

### DIFF
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -184,9 +184,9 @@ class ServerHandler extends ServiceCall {
     _service = service!;
     _descriptor = descriptor;
 
-    final error = await _applyInterceptors();
+    final (error, trace) = await _applyInterceptors();
     if (error != null) {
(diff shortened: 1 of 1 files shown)


T0bst3r:
I have signed the document at least 4 times by now, it always tells me that it's accepted it and redirects me to my other pull request (#667) but the status here does not change. Emails of both PRs are identical. What can I do?

devoncarew:
> I have signed the document at least 4 times by now, it always tells me that it's accepted it and redirects me to my other pull request (#667) but the status here does not change. Emails of both PRs are identical. What can I do?

As a drive by comment, does the CLA comment above provide useful debugging info? It says:

> The email address for the commit is not linked to the GitHub account, preventing the EasyCLA check.

And I see from the patch for this PR that the email address its checking against is `leuteltt+git@gmail.com`.

Issue #669, opened 2023-09-18
GRPC server seems to always compress messages even if client does not support it.


RubenGarcia:
<short description>
GRPC server seems to always compress messages even if client does not support it. See
https://github.com/Cysharp/YetAnotherHttpHandler/issues/19
where client is complaining with the line
Request sent 'identity' grpc-encoding value with compressed message. #19

<version of the grpc-dart packages used; see your `pubspec.lock` file>
  grpc:
    dependency: "direct main"
    description:
      name: grpc
      sha256: "220ffb2218288f4e7dea487242e08b9c6277596d9b6f3f10ba50be96771a032d"
      url: "https://pub.dev"
    source: hosted
    version: "3.2.3"

## Repro steps

1. Generate a dart grpc server
2. Use Unity and YetAnotherHttpHandler to connect to it

Expected result:
The client receives the correct answer

Actual result: 
Request sent 'identity' grpc-encoding value with compressed message. #19

RubenGarcia:
Removing compression using
```
 codecRegistry: CodecRegistry(codecs: const [/*GzipCodec(), */IdentityCodec()]),
```
does not make a difference.

RubenGarcia:
Version: "3.2.4" is also affected.

RubenGarcia:
Tested on android and on osx, both platforms have the issue.

RubenGarcia:
Closing as fix was merged.

PR #692, opened 2024-03-19
Fix https://github.com/grpc/grpc-dart/issues/669

RubenGarcia:
If message-encoding is set to identity, then compression flag should be 0.


RubenGarcia:
Correct merge request with EasyCLA is
https://github.com/grpc/grpc-dart/pull/693

PR #693, opened 2024-03-19
Fix issue 669

RubenGarcia:
https://grpc.github.io/grpc/core/md_doc_compression.html indicates that compression flag should not be set if grpc-encoding is identity.

This PR makes grpc-dart set the compression flag properly in this case.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -19,7 +19,7 @@ jobs:
         sdk: [3.0.0, dev]
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
-      - uses: dart-lang/setup-dart@b64355ae6ca0b5d484f0106a033dd1388965d06d
+      - uses: dart-lang/setup-dart@fedb1266e91cf51be2fdb382869461a434b920a3
         with:
           sdk: ${{ matrix.sdk }}
       - name: Report version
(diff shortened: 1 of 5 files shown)


mosuem:
Thanks a lot for the investigation and the fix! If you could add a changelog entry describing your change (with a link to this PR), as well as adding the information from https://github.com/grpc/grpc-dart/issues/669#issuecomment-2006654280 to this PR description, that would help a lot.

RubenGarcia:
> Thanks a lot for the investigation and the fix! If you could add a changelog entry describing your change (with a link to this PR), as well as adding the information from [#669 (comment)](https://github.com/grpc/grpc-dart/issues/669#issuecomment-2006654280) to this PR description, that would help a lot.

Done and fixed the two issues with the workflow.

RubenGarcia:
I´m getting this error in the other workflow:
```
Validating package:grpc 3.2.4 (dir=/home/runner/work/grpc-dart/grpc-dart):grpc
pubspec:
  - version: 3.2.4
changelog:
* Set compressed flag correctly for grpc-encoding = identity. Fixes [#669](https://github.com/grpc/grpc-dart/issues/669) (https://github.com/grpc/grpc-dart/pull/693)
error: pubspec version (3.2.4) and changelog (null) don't agree
```
Since 3.2.4 has already released, I will change both to 3.2.5.

mosuem (changes requested):
Looking good! If you could add a small test for this to not risk having regressions, we can merge.

RubenGarcia:
> Looking good! If you could add a small test for this to not risk having regressions, we can merge.

Test added.

RubenGarcia:
The error when running dart / test (windows-latest) when running
dart run --enable-vm-service --timeline-streams=Dart test/timeline_test.dart
Could not find a command named "/hostedtoolcache/windows/dart/3.4.0-260.0.d...
seem to be unrelated to my changes.

mosuem:
Right, I will take a look.

RubenGarcia:
Is there anything else needed here?

RubenGarcia:
The failed tests complaining about
dart:js_interop_unsafe
seem to be independent of my changes, although they are a transitive dependency of one of my imports.

PR #722, opened 2024-07-03
fix: keep alive timeout finishes transport instead of connection shutdown

steffenhaak:
Currently, when the keep alive manager runs into a timeout, the connection is shutdown. The channel, however, cannot recover and stays in a stuck state.

Instead, we should finish the transport, which eventually will change the connection to be in transient failure mode, from where it recovers, once a stable connection can be established again.

### DIFF
--- a/lib/src/client/http2_connection.dart
+++ b/lib/src/client/http2_connection.dart
@@ -113,7 +113,7 @@ class Http2ClientConnection implements connection.ClientConnection {
               transport.ping();
             }
           },
-          onPingTimeout: () => shutdown(),
+          onPingTimeout: () => transport.finish(),
         );
         transport.onFrameReceived
             .listen((_) => keepAliveManager?.onFrameReceived());


mosuem:
Thanks for the PR! Sounds good at first glance, but I will check the [specs](https://github.com/grpc/grpc/blob/master/doc/keepalive.md) and take a look at this PR next week.

steffenhaak:
Quote from the specs:

> If a keepalive ping is not blocked and is sent on the transport, then the keepalive watchdog timer is started which will close the transport if the ping is not acknowledged before it fires.

Closing the transport should be the appropriate thing to do. But take your time and think about it thoroughly.

I'm using a modified version for now, so there's absolutely no pressure on my side. Please let me know if I can assist you in any way with this.

mosuem:
LGTM in general, I filed a PR on your fork to address the failing test. Let me know what you think.

mosuem:
@steffenhaak Also sorry for the delay!

steffenhaak:
@mosuem Great, thank you! I will take a look at it, most likely by the end of this week, and do the rebasing...

steffenhaak:
@mosuem Your changed test looks good to me. I rebased the PR on the current main branch. So I guess this PR is ready to be merged :)

mosuem:
> @mosuem Your changed test looks good to me. I rebased the PR on the current main branch. So I guess this PR is ready to be merged :)

It seems to need a changelog entry, but then yes!

steffenhaak:
@mosuem done :)

PR #724, opened 2024-07-17
Use `Map.of` instead of `Map.from` in grpc client.

lrhn:
`Map.of` creates a new map with the same keys, values and *type* as the original map,
when used without type arguments or context type, where `Map.from` creates a `Map<dynamic, dynamic>`.
(This code failed on an attempt to make `Map.unmodifiable` be more strictly typed,
like `Map.of` instead of `Map.from`, showing that an intermediate map had type
`Map<dynamic, dynamic>` unnecessarily).

Same for using `List.of` instead of `List.from`.

The new code should be (microscopically) more efficient and type safe, and is forwards-compatible
with a stronger type on `Map.unmodifiable`.

(The code can be optimized more. For example
`List.of(list1)..addAll(list2)` can be just `list1 + list2` or `[...list1, ...list2]`,
both of which may know the total number of elements when doing the
initial list allocation. This is a minimal change to allow the type changes
for `.unmodifiable` to get past this very initial blocker in internal tests.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.2
+
+* Internal optimization to client code.
+
 ## 4.0.1
 
 * Fix header and trailing not completing if the call is terminated. Fixes [#727](https://github.com/grpc/grpc-dart/issues/727)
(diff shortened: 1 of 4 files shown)


mosuem (approved):

mosuem:
@lrhn Can these changes be merged?

lrhn:
Well, now it has conflicts! But otherwise yes. Fixing conflicts.

Issue #727, opened 2024-08-08
Awaiting ResponseFuture.headers hangs indefinetly for not implemented services


c-lucera-pvotal:
<short description>

Reproduced in `grpc: 4.0.0`

Per [documentations](https://pub.dev/documentation/grpc/latest/grpc/ResponseFuture/headers.html
), we should be able to await the headers in any give ResponseFuture call.

However when the call is not implemented, the code hangs indefinitely without completing.

### Sample
Using the following code I expect to always have a print at some point
```dart
Future<void> testService() async {
    try {
      final client = MyService.instance.client;
      final request = TestRequest(name: "Test");
      final responseFuture = client.sayHello(request);
    
      var header = await responseFuture.headers; //ISSUE IN THIS LINE
      var helloResponse = await responseFuture;
    
      print(helloResponse.message);
    } catch (e) {
      print(e);
    }
}
```
But in reality what is happening is that the `await responseFuture.headers;` never completes, hanging this future indefinitely

I created [this repository ](https://github.com/c-lucera-pvotal/grpc_headers_test) to test this case, it defines a sample proto and it already include the generated pb files inside `lib/services`, also it includes a Simple application that shows a button to call the service and await for the result.

### How to test
1. Clone [this repo](https://github.com/c-lucera-pvotal/grpc_headers_test)
2. On the root directory run the project, by using `flutter run` on the console for your device of choice
3. Tap on the `Send request` button.

#### Expected result
A loader is briefly shown, then a error is shown in the text over the button indicating the service is not available.

#### Actual result
The loader never goes away because the `var header = await responseFuture.headers;` at line 59 of main.dart never completes.

## Details
Running on web I get the following log, showing that something failed asyncronously but it is not rethrown/catched
```
Error
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 329:10  createErrorWithStack
dart-sdk/lib/_internal/js_dev_runtime/patch/core_patch.dart 265:28            _throw
dart-sdk/lib/core/errors.dart 120:5                                           throwWithStackTrace
dart-sdk/lib/async/zone.dart 1386:11                                          callback
dart-sdk/lib/async/schedule_microtask.dart 40:11                              _microtaskLoop
dart-sdk/lib/async/schedule_microtask.dart 49:5                               _startMicrotaskLoop
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 181:7            <fn>
```

[...]

PR #728, opened 2024-08-08
fix: fix headers not completing when call is terminated

c-lucera-pvotal:
This PR attempt to solve the issue with headers not being correctly completed after a call is closed as pointed out in https://github.com/grpc/grpc-dart/issues/727

To fix the issue I have added in the `_terminate()` method a check to complete with errors any completer still going (I found `_headers` and `_trailers`)

### DIFF
--- a/test/keepalive_test.dart
+++ b/test/keepalive_test.dart
@@ -49,7 +49,7 @@ void main() {
       services: [FakeEchoService()],
       keepAliveOptions: serverOptions,
     );
-    await server.serve(address: 'localhost', port: 8082);
+    await server.serve(address: 'localhost', port: 0);
     fakeChannel = FakeClientChannel(
       'localhost',
       port: server.port!,
--- a/test/keepalive_test.dart
+++ b/test/keepalive_test.dart
@@ -49,7 +49,7 @@ void main() {
       services: [FakeEchoService()],
       keepAliveOptions: serverOptions,
     );
-    await server.serve(address: 'localhost', port: 8081);
+    await server.serve(address: 'localhost', port: 8082);
     fakeChannel = FakeClientChannel(
       'localhost',
       port: server.port!,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.0.1
+
+* Fix header and trailing not completing if the call is terminated. Fixes [#727](https://github.com/grpc/grpc-dart/issues/727)
+
 ## 4.0.0
 
 * Set compressed flag correctly for grpc-encoding = identity. Fixes [#669](https://github.com/grpc/grpc-dart/issues/669) (https://github.com/grpc/grpc-dart/pull/693)
--- a/lib/src/client/call.dart
+++ b/lib/src/client/call.dart
@@ -484,12 +484,10 @@ class ClientCall<Q, R> implements Response {
       futures.add(_responseSubscription!.cancel());
     }
     if (!_headers.isCompleted) {
-      _headers.completeError(
-          GrpcError.unimplemented('Request terminated before headers'));
+      _headers.complete({});
     }
     if (!_trailers.isCompleted) {
-      _trailers.completeError(
-          GrpcError.unimplemented('Request terminated before trailers'));
+      _trailers.complete({});
     }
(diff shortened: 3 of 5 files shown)


r-durao-pvotal:
@kevmoo who would be the right person to take a look at this PR? Thanks in advance!

mraleph:
Please add a regression test and then we can land it.

c-lucera-pvotal:
@mraleph 
Added test and fixed CI checks, both for CHANGELOG and analyze

c-lucera-pvotal:
@mraleph 
Added the requested test and fixed CI issues like Changelog notes.

However I had to change the previous logic from throwing an error
```dart
    if (!_headers.isCompleted) {
      _headers.complete({});
    }
```

To completing with an empty value
```dart
    if (!_headers.isCompleted) {
      _headers.completeError(GrpcError.unimplemented('Request terminated before headers'));
    }
```

The reason about the change was that the error throwing was really distruptive for the current implementations, instead of receving a error only on active listeners (I.e whoever was actively looking at that completer) all the cancelled calls received 3 throws, one for the cancellation, one for the headers not completed and one for the trailers not completed, this meant that all the tests in this project were failing due to the not catched exceptions, and at the same time, anyone cancelling the call had to deal with 3 different throws, I hope this change make sense to you.

c-lucera-pvotal:
@mraleph saw some failing tests, but those seems unrelated to my changes, any hint about those?

mosuem:
These keepalive tests can be flaky, but this doesn't look that way. Let me check.

c-lucera-pvotal:
It looks like that the tearDown somehow is not awaiting the closing before the next setUp is running, maybe they are running in parallel?

in [line 52 of keep_alive_test.dart,](https://github.com/grpc/grpc-dart/blob/c18e185bb09d94a6760e4c43c9bf8c05c0f0e4c9/test/keepalive_test.dart#L52C5-L52C58) if you use a different port for each serve you will see the tests not failing anymore

```dart
//before tests setup
int port = 8081;

//Line 52
 await server.serve(address: 'localhost', port: port++);
 ```
 
 Want me to add this?

**Comment by c-lucera-pvotal** [comment]
Another thing I discovered, if I run only keep alive tests it works out of the box, so probably tests are running in parallel with other tests using the port 8081, another fix is to use the port 8282 on line 52

```dart
/Line 52
 await server.serve(address: 'localhost', port: 8082);
 ```

**Comment by c-lucera-pvotal** [comment]
Basically those two tests
https://github.com/grpc/grpc-dart/blob/master/test/keepalive_test.dart
https://github.com/grpc/grpc-dart/blob/master/test/server_cancellation_test.dart

uses port 8081, this is not an issue when run one by one, but running the whole suite with `flutter test` causes he keep_alive failing because the server cancellation test run and find the port occupied, I can fix this in my pr by changing the port if you want

**Comment by c-lucera-pvotal** [comment]
Pushed the fix for `keepalive_test.dart` let me know if you prefer to revert it and solve in a different way

File: test/keepalive_test.dart [hunk 1]
```diff
@@ -49,7 +49,7 @@ void main() {
       services: [FakeEchoService()],
       keepAliveOptions: serverOptions,
     );
-    await server.serve(address: 'localhost', port: 8081);
+    await server.serve(address: 'localhost', port: 8082);
```
kevmoo, line 52: What's w/ this change? Necessary?

mosuem: This fixes the flaky keepalive tests.

c-lucera-pvotal: I think 0 is fine, but you need some consistency along tests, I can pass 0 in the failing one, but you should consider another PR to fix all the random port passed around (unless they are really needed) 
image

Testing with 0, if the tests passes I will push the change for keepalive_test

c-lucera-pvotal: @mraleph pushed `keepalive_test.dart` with port 0

mraleph: Thanks! We should definitely change all the tests which don't need a fixed port to use `0` as a port. This kinda slipped under the radar in reviews. We'll do it in another PR.

mraleph (approved):

Issue #715, opened 2024-05-23
Migrate off legacy JS/HTML apis


minoic:
Wasm support is now stable since flutter 3.22 published, i got UNAVAILABLE error while testing my web app with wasm.

Version of the grpc-dart packages used: v3.2.4 and grpc/grpc-dart master branch.

## Repro steps

1. Create a grpc channel using `GrpcOrGrpcWebClientChannel.toSeparateEndpoints`, write some request code.
2. Build: `flutter build web --wasm`.
3. To run a Flutter app that has been compiled to Wasm, follow [Support for WebAssembly (Wasm)](https://docs.flutter.dev/platform-integration/web/wasm).

Expected result: The request successfully sent as canvaskit mode does.

Actual result: 

```
gRPC Error (code: 14, codeName: UNAVAILABLE, message: Error connecting: Unsupported operation: SecurityContext constructor, details: null, rawResponse: null, trailers: {})
```

## Details

This error could be caused by two points: 

1. Wrong package imported in `grpc_or_grpcweb.dart`. Because of wasm has no html package, conditional import recognized wasm as non-web platform.
2. Once correct platform recognized, `xhr_transport.dart` will be used to transport data, which uses html package, but "dart:html is being replaced with [package:web](https://pub.dev/packages/web). Package maintainers should migrate to package:web as soon as possible to be compatible with Wasm. Read the [Migrate to package:web](https://dart.dev/interop/js-interop/package-web) page for guidance."

I tried to fix it in [https://github.com/minoic/grpc-dart](https://github.com/minoic/grpc-dart), it worked in my case but more test should be performed later.

mosuem:
Thanks for reporting this, we should migrate `package:grpc` off `dart:html`. cc @kevmoo

r-durao-pvotal:
Is there any ETA on this @mosuem , @kevmoo ?

kevmoo:
Everyone is busy. Happy to accept a pull request!

zs-dima:
+1
Error: /home/flutter/.pub-cache/hosted/pub.dev/grpc-3.2.4/lib/src/client/transport/xhr_transport.dart:17:8: Error: Dart library 'dart:html' is not available on this platform.

minoic:
> +1 Error: /home/flutter/.pub-cache/hosted/pub.dev/grpc-3.2.4/lib/src/client/transport/xhr_transport.dart:17:8: Error: Dart library 'dart:html' is not available on this platform.

[...]

PR #730, opened 2024-08-17
update: Migrate off legacy JS/HTML apis

minoic:
The purpose of this PR is to fix the issue mentioned in #715, I have test both unary/server side stream requests on canvaskit/wasm platform.Thanks to @hyunw55 `s comment in https://github.com/grpc/grpc-dart/issues/715#issuecomment-2133525797.

Main updates include:

1. add web v1.0.0 package and `XMLHttpRequest` to replace html package.
2. use `dart.library.io` to determine the current platform.

### DIFF
--- a/lib/grpc_or_grpcweb.dart
+++ b/lib/grpc_or_grpcweb.dart
@@ -13,8 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import 'src/client/grpc_or_grpcweb_channel_grpc.dart'
-    if (dart.library.html) 'src/client/grpc_or_grpcweb_channel_web.dart';
+import 'src/client/grpc_or_grpcweb_channel_web.dart'
+    if (dart.library.io) 'src/client/grpc_or_grpcweb_channel_grpc.dart';
 import 'src/client/http2_channel.dart';
 import 'src/client/options.dart';
 
--- a/lib/grpc_or_grpcweb.dart
+++ b/lib/grpc_or_grpcweb.dart
@@ -13,8 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import 'src/client/grpc_or_grpcweb_channel_web.dart'
-    if (dart.library.io) 'src/client/grpc_or_grpcweb_channel_grpc.dart';
+import 'src/client/grpc_or_grpcweb_channel_grpc.dart'
+    if (dart.library.js_interop) 'src/client/grpc_or_grpcweb_channel_web.dart';
 import 'src/client/http2_channel.dart';
 import 'src/client/options.dart';
 
--- a/lib/src/client/transport/xhr_transport.dart
+++ b/lib/src/client/transport/xhr_transport.dart
@@ -59,12 +59,14 @@ class XhrTransportStream implements GrpcTransportStream {
       if (_incomingProcessor.isClosed) {
         return;
       }
-      // TODO: dart-lang/web#285 use 'if' for now
-      if (_request.readyState == XMLHttpRequest.HEADERS_RECEIVED) {
-        _onHeadersReceived();
-      } else if (_request.readyState == XMLHttpRequest.DONE) {
-        _onRequestDone();
-        _close();
+      switch (_request.readyState) {
+        case XMLHttpRequest.HEADERS_RECEIVED:
+          _onHeadersReceived();
+          break;
+        case XMLHttpRequest.DONE:
+          _onRequestDone();
+          _close();
+          break;
       }
     });
 
@@ -99,7 +101,7 @@ class XhrTransportStream implements GrpcTransportStream {
   bool _validateResponseState() {
     try {
       validateHttpStatusAndContentType(
-          _request.status, _parseHeaders(_request.getAllResponseHeaders()),
+          _request.status, _request.responseHeaders,
           rawResponse: _request.responseText);
       return true;
     } catch (e, st) {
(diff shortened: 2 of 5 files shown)


lib/src/client/transport/xhr_transport.dart
```diff
@@ -45,23 +46,22 @@ class XhrTransportStream implements GrpcTransportStream {
...
+        .listen((data) => _request.send(Int8List.fromList(data).toJS), cancelOnError: true, onError: _onError);
 
-    _request.onReadyStateChange.listen((data) {
+    _request.onReadyStateChange.listen((_) {
       if (_incomingProcessor.isClosed) {
         return;
       }
       switch (_request.readyState) {
-        case HttpRequest.HEADERS_RECEIVED:
+        case 2:
```
aran, line 61: It looks like the constants are available in package:web:

https://github.com/dart-lang/web/blob/4996dc2acd30ff06f7e500ec76fde8a66db82c0c/lib/src/dom/xhr.dart#L94-L98

lib/grpc_or_grpcweb.dart
```diff
@@ -13,8 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import 'src/client/grpc_or_grpcweb_channel_grpc.dart'
-    if (dart.library.html) 'src/client/grpc_or_grpcweb_channel_web.dart';
```
aran, line 17: You can avoid the swap and use dart.library.js_interop in place of dart.library.html. There's some documentation here: https://dart.dev/interop/js-interop/package-web that has an example of that.

lib/src/client/transport/xhr_transport.dart
```diff
@@ -115,17 +108,15 @@ class XhrTransportStream implements GrpcTransportStream {
     if (!_validateResponseState()) {
       return;
     }
-    _incomingMessages.add(GrpcMetadata(_request.responseHeaders));
```
aran, line 118: It looks like package:http exposes an extension method on XmlHttpRequest to make `responseHeaders` available:

https://github.com/dart-lang/http/blob/b97b8dc22ea808c4fbd63f73abd7af8ecf694323/pkgs/http/lib/src/browser_client.dart#L124

Using it would avoid needing to implement HTTP header parsing below in _parseHeaders.

aran:
@kevmoo do you happen to know who would be the right maintainers to look at this?

kevmoo:
Thanks for chiming in here @aran 
And thanks for the PR @minoic 

I'll work to find a reviewer!

pubspec.yaml
kevmoo, line 1: You'll need to bump the min SDK to Dart 3.4 – please update the version to v4.0.1 and also add an entry to the changelog

lib/src/client/transport/xhr_transport.dart
minoic: I tried to write `case XMLHttpRequest.HEADERS_RECEIVED:` but i got an error from IDEA: 

> The expression of a constant pattern must be a valid constant. 
> Try making the expression a valid constant.

and compile error:

```
../grpc-dart/lib/src/client/transport/xhr_transport.dart:61:29: Error: Constant evaluation error:
        case XMLHttpRequest.HEADERS_RECEIVED:
                            ^
../grpc-dart/lib/src/client/transport/xhr_transport.dart:61:29: Context: The invocation of 'XMLHttpRequest|HEADERS_RECEIVED' is not allowed in a constant expression.
        case XMLHttpRequest.HEADERS_RECEIVED:
                            ^
../grpc-dart/lib/src/client/transport/xhr_transport.dart:61:29: Context: While analyzing:
        case XMLHttpRequest.HEADERS_RECEIVED:
                            ^
Failed to compile application.
```

lib/src/client/transport/xhr_transport.dart
minoic: This extension method seems to be implicity and private, so if a name is given to this extension i may use `import 'package:http/src/browser_client.dart'` and `_request.responseHeaders`.

https://stackoverflow.com/questions/59092361/import-extension-method-from-another-file-in-dart

lib/src/client/transport/xhr_transport.dart
mraleph: @srujzs any specific reason these are generated as external getters rather than constants? makes it impossible to use them as constants in a switch.

srujzs: The only benefit to keeping them `external` is in case the browser diverges from the IDL specification, but I expect them to rarely ever happen. This came up recently elsewhere and it's better if we emit them as Dart values to support switches. I filed https://github.com/dart-lang/web/issues/285.

mosuem: @minoic I would still use the provided values, even if not constants, and just not use a switch-case for now. You could leave a todo note pointing to that bug in dart-lang/web.

lib/src/client/transport/xhr_transport.dart
mosuem: I filed https://github.com/dart-lang/web/pull/298

mosuem (changes requested):

lib/src/client/transport/xhr_transport.dart
srujzs: The ability to switch over them has landed and should be in `1.1.0`.

kevmoo:
We published a new `pkg:web` with those HTTP helpers

We need to rebase this fix, too

kevmoo:
looks like we need to do another merge here

aran:
@kevmoo see #747 for an attempt to push this forward.

mosuem:
Closing in favor of #747

PR #748, opened 2024-12-04
fix: Updates the grpc-web example to avoid dart:html

aran:
Updates the grpc-web example to avoid dart:html.


mosuem (approved):
Good to land after #747

aran:
Will be based on #750 instead of #747

aran:
I think we should label this `skip-changelog-check`, because there is already a relevant changelog entry: 

    * Upgrade `example/grpc-web` code.
    
Are you ok with that @mosuem ? (I'm unable to apply that label).

mosuem (approved):

PR #749, opened 2024-12-04
fix: Use package:web to get HttpStatus

aran:
Removes HttpStatus from io_bits, thereby breaking the dependency on dart:html.

Code that previously used HttpStatus from dart:html or dart:io now gets it from package:web.

Alternatively, this could instead use HttpStatus from dart:io on platforms with dart:io and get it from package:web only otherwise. If that's preferred we can do it that way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Upgrade to `package:lints` version 5.0.0 and Dart SDK version 3.5.0.
 * Upgrade `example/grpc-web` code.
 * Update xhr transport to migrate off legacy JS/HTML apis.
+* Use `package:web` to get `HttpStatus`
 
 ## 4.0.1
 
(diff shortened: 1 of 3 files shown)


mosuem:
> Code that previously used HttpStatus from dart:html or dart:io now gets it from package:web.

You mean the user has to import that himself, right?
 
> Alternatively, this could instead use HttpStatus from dart:io on platforms with dart:io and get it from package:web only otherwise. If that's preferred we can do it that way.

At first glance that sounds preferable and less breaking.

aran:
> At first glance that sounds preferable and less breaking.

Switched.

mosuem:
LGTM, but please add a changelog and rev the version so that the health check stops complaining :)

aran:
@mosuem should be good to go here (?)

Issue #756, opened 2025-01-29
update use of deprecated dart:html apis


devoncarew:
We currently have 5 analysis issues due to use of (deprecated) dart:html apis.

PR #757, opened 2025-01-31
use package:web for the grpc-web example

devoncarew:
- use package:web for the grpc-web example
- some work towards #756


devoncarew:
Hmm, it looks like there are already some in-flight PRs to address #756 (https://github.com/grpc/grpc-dart/pulls); we should probably prefer to land some combination of them.

aran:
I think this is covered by #748 as well but doesn't really matter which one lands.

mosuem:
Closing in favor of #748

Issue #611, opened 2023-03-24
CORS not supported on a Dart server


Levi-Lesches:
<short description>

When using `package:grpc` on a Dart server, you make a `Server` and then call `server.serve()`. This server cannot handle CORS requests in the way `package:shelf` can, and so all requests from the Flutter side are broken.

grpc version: `v3.1.0`

## Repro steps

```dart
// LoginService is a regular rpc service
void main() async {
  final server = Server([LoginService()]);
  await server.serve(port: 9000);
  print("Server running on http://localhost:9000");
}
```

Expected result: Sending requests to this server works

Actual result: 
```
gRPC Error (code: 2, codeName: UNKNOWN, message: HTTP request completed without a status (potential CORS issue)
```

mraleph:
Note that the server implementation only talks gRPC native protocol, does not support gRPC-Web clients directly. Hence no CORS support or anything like that. You need a proxy in between which would convert gRPC-Web to gRPC.

Levi-Lesches:
Yes, I saw that elsewhere. Would it be possible to provide, in this package or another, native Dart support for `gRPC-Web` without the need for a separate proxy?

mraleph:
It is definitely possible and I think we will accept a PR implementing gRPC-Web server.

Issue #591, opened 2022-12-03
[Feature Request] server-side "wrapper" interceptors


cbatson:
Hello, I'd like to see something like `Interceptor`s on the server side, but more akin to the [UnaryInterceptor of gRPC-Web](https://grpc.io/blog/grpc-web-interceptor/). This is something I have appreciated from other server frameworks. This style of interceptor would "wrap" around the service call, which yields the existing functionality of `Interceptor`s (such as vetoing a request) and more. Such "wrapping" interceptors would be able to manipulate both the incoming request as well as the outgoing response.

There are several interesting use cases for this. One such use case would be instrumenting all service calls in a uniform way.

```
Interceptor 1
|   Interceptor 2
|   |   Interceptor 3
|   |   |   service call
|   |   Interceptor 3
|   Interceptor 2
Interceptor 1
```

mraleph:
I think it makes sense to deprecate current interceptor functionality and align client and server interceptor APIs. Feel free to contribute PR with the implementation.

cbatson:
Note to self, consider addressing #546 when implementing this.

Levi-Lesches:
A server-side interceptor could possibly add CORS headers to a response, fixing #611

PR #762, opened 2025-03-03
Add server interceptor acting as a middleware

ThHareau:
Fix for #591

Example of usage: 

```dart
class SentryInstrumenterInterceptor extends $grpc.ServerInterceptor {
  @override
  Stream<R> interceptStreaming<Q, R>(
      $grpc.ServiceCall call,
      $grpc.ServiceMethod<Q, R> method,
      Stream<Q> requests,
      $grpc.ServerStreamingInvoker<Q, R> invoker) async* {
    final name = method.name;
    final transaction = Sentry.startTransaction(method.name, 'grpc');
    print('Before $name');

    yield* super.interceptStreaming(call, method, requests, invoker);

    print('After $name');
    transaction.finish();
  }
}
```

--- 

I did not check if this PR solves #611, anyway if it does not it would probably be nice to open a second PR, this one is complex enough

### DIFF
--- a/lib/src/server/interceptor.dart
+++ b/lib/src/server/interceptor.dart
@@ -27,3 +27,18 @@ import 'service.dart';
 /// If the interceptor returns null, the corresponding [ServiceMethod] of [Service] will be called.
 typedef Interceptor = FutureOr<GrpcError?> Function(
     ServiceCall call, ServiceMethod method);
+
+typedef ServerStreamingInvoker<Q, R> = Stream<R> Function(
+    ServiceCall call, ServiceMethod<Q, R> method, Stream<Q> requests);
+
+/// A gRPC Interceptor.
+///
+/// An interceptor is called around the corresponding [ServiceMethod] invocation.
+/// If the interceptor throws [GrpcError], the error will be returned as a response. [ServiceMethod] wouldn't be called if the error is thrown before calling the invoker.
+/// If the interceptor modifies the provided stream, the invocation will continue with the provided stream.
+abstract class ServerInterceptor {
+  Stream<R> intercept<Q, R>(ServiceCall call, ServiceMethod<Q, R> method,
+      Stream<Q> requests, ServerStreamingInvoker<Q, R> invoker) {
+    return invoker(call, method, requests);
+  }
+}
--- a/lib/src/server/service.dart
+++ b/lib/src/server/service.dart
@@ -76,8 +76,12 @@ class ServiceMethod<Q, R> {
     var invoker = _createCall();
 
     for (final interceptor in interceptors.reversed) {
+      final delegate = invoker;
+      // invoker is actually reassigned in the same scope as the above function,
+      // reassigning invoker in delegate is required to avoid an infinite
+      // recursion
       invoker = (call, method, requests) =>
-          interceptor.intercept<Q, R>(call, method, requests, invoker);
+          interceptor.intercept<Q, R>(call, method, requests, delegate);
     }
 
     return invoker(call, this, requests);
--- a/lib/src/server/service.dart
+++ b/lib/src/server/service.dart
@@ -17,6 +17,7 @@ import 'dart:async';
 
 import '../shared/status.dart';
 import 'call.dart';
+import 'interceptor.dart';
 
 /// Definition of a gRPC service method.
 class ServiceMethod<Q, R> {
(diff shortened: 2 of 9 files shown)


lib/src/server/interceptor.dart
```diff
@@ -27,3 +27,18 @@ import 'service.dart';
...
+    ServiceCall call, ServiceMethod<Q, R> method, Stream<Q> requests);
+
+/// A gRPC Interceptor.
+///
+/// An interceptor is called around the corresponding [ServiceMethod] invocation.
+/// If the interceptor throws [GrpcError], the error will be returned as a response. [ServiceMethod] wouldn't be called if the error is thrown before calling the invoker.
+/// If the interceptor modifies the provided stream, the invocation will continue with the provided stream.
+abstract class ServerInterceptor {
+  Stream<R> intercept<Q, R>(ServiceCall call, ServiceMethod<Q, R> method,
+      Stream<Q> requests, ServerStreamingInvoker<Q, R> invoker) {
```
ThHareau, line 41: The goal here is to be close to the Client implementation. As the handler is a Stream, I did not add unary. It is probably doable tho.

example/grpc-web/lib/app.dart
```diff
@@ -61,7 +61,7 @@ class EchoApp {
         ..append(HTMLSpanElement()
           ..classList.add('label')
           ..classList.addAll(cssClass)
-          ..text = message)));
+          ..textContent = message)));
```
ThHareau, line 64: Should I keep this? CI was failing here, even if this has nothing to do with this PR

mosuem: This is fixed in #765

ThHareau: Okk, I'll revert as soon as 765 is merged.

ThHareau:
Should I do anything about the changelog health check?

mosuem:
> Should I do anything about the changelog health check?

No, the breaking change should not be regarded as breaking and is from a different PR (note to self: the check should show that the breakage was introduced elsewhere..). The version is probably fine. 

I'll take a closer look at this once I get to it, quite busy ATM unfortunately. But thanks a lot already for the contribution!

ThHareau:
> > Should I do anything about the changelog health check?
> 
> No, the breaking change should not be regarded as breaking and is from a different PR (note to self: the check should show that the breakage was introduced elsewhere..). The version is probably fine.
> 
> I'll take a closer look at this once I get to it, quite busy ATM unfortunately. But thanks a lot already for the contribution!

No worries, it can wait a bit :slightly_smiling_face:  Just fyi, I'll be away from the 12th to end of march, so I won't be able to answer during this period. But I'll reply afterward.

ThHareau:
CI should pass now :)

ThHareau:
@mosuem  Is the changelog failure something I can act on?

lib/src/server/service.dart
```diff
@@ -48,19 +49,39 @@ class ServiceMethod<Q, R> {
...
+
+  Stream<R> handle(
+    ServiceCall call,
+    Stream<Q> requests,
+    List<ServerInterceptor> interceptors,
+  ) {
+    var invoker = _createCall();
+
+    for (final interceptor in interceptors.reversed) {
+      final delegate = invoker;
```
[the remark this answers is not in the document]
ThHareau: It comes from these lines https://github.com/grpc/grpc-dart/blob/f5508d9801dd52c1bc9929d0c5b669d25ad2fc3f/lib/src/client/client.dart#L46-L58

I suppose it was done to improve readability of the recursive call on the line below: 
```dart
      final delegate = invoker;
      invoker = (call, method, requests) =>
          interceptor.intercept<Q, R>(call, method, requests, delegate);
```
vs
```dart
      invoker = (call, method, requests) =>
          interceptor.intercept<Q, R>(call, method, requests, invoker);
```

I don't have a strong opinion about either solution, just tell me which version to keep :slightly_smiling_face:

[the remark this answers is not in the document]
mosuem: Ah ok. Me neither, but if you add a local variable I would add some documentation at least. Or just inline for simplicity :)

Then it's LGTM!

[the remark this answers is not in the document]
ThHareau: Let's go for simplicity then!

ThHareau:
@mosuem any idea on how to make the CI pass? I see that both checks failed with error `Error: Process completed with exit code 143.`, which corresponds to a `SIGTERM` if I understood correctly. Could you try relaunching it?

mosuem:
Locally, the tests hang as well, which can be helped by commenting out [serverInterceptors: serverInterceptors..insert(0, serverInterceptor),](https://github.com/grpc/grpc-dart/blob/e6458841c83ba3d4d29cc82982834deeb6fa0cfb/test/src/server_utils.dart#L167) - which seems to be the culprit.

ThHareau:
Hi @mosuem , do you think this could get merged? Do you have any further review? 

Thanks :slightly_smiling_face:

mosuem (approved):

mosuem:
Thank you a lot, and sorry for the delay!

PR #775, opened 2025-03-31
Allow the latest pkg:googleapis_auth

kevmoo:
Closes https://github.com/grpc/grpc-dart/pull/772

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   async: ^2.5.0
   crypto: ^3.0.0
   fixnum: ^1.0.0
-  googleapis_auth: ">=1.0.0 <3.0.0"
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   async: ^2.5.0
   crypto: ^3.0.0
   fixnum: ^1.0.0
-  googleapis_auth: ^1.1.0
+  googleapis_auth: ">=1.0.0 <3.0.0"
```
mosuem, line 18: Why leave the dependency on the old `googleapis_auth`?

kevmoo: To let folks upgrade. There are no breaking changes as far as this package is concerned.

kevmoo:
PTAL?

mraleph (approved):

PR #786, opened 2025-05-26
Fix CallOptions.mergeWith

mraleph:
It should have symmetric behavior: given `WebCallOptions a` and `CallOptions b` both `a.mergeWith(b)` and `b.mergeFrom(a)` should return `WebCallOptions`.


devoncarew:
FYI that this release hasn't been published. For people with commit access, it can be published from the link here: https://github.com/grpc/grpc-dart/actions/runs/15252145193.

PR #790, opened 2025-06-09
update deps; simplify analysis options

devoncarew:
- require Dart 3.6
- update deps for the example packages
- simplify analysis options for the example packages


mosuem (approved):

pubspec.yaml
```diff
@@ -9,24 +9,24 @@ topics:
   - rpc
 
 environment:
-  sdk: ^3.5.0
+  sdk: ^3.6.0
```
mosuem, line 12: We could go to `3.8.0` as the current stable is beyond that.. And then run a `dart pub upgrade --tighten`, as the lowest bounds don't build for me locally.

mosuem (approved):

pubspec.yaml
devoncarew: sgtm; will do

devoncarew:
FYI, moving the 3.8 required a full dart format of all dart source files.

mosuem (approved):

PR #792, opened 2025-06-09
export additional generated protobuf symbols

devoncarew:
- export additional generated protobuf symbols (`Any`)

From the health check, we're referencing symbols from the public api that are not exported. This exports some of those symbols.

```
The following packages contain symbols visible in the public API, but not exported by the library.
Export these symbols or remove them from your publicly visible API.

Package	Leaked API symbols
grpc	$1.Duration
    	Any
```


devoncarew:
It looks like `Duration` is leaked from `package:grpc/src/generated/google/rpc/error_details.pb.dart`, `RetryInfo.retryDelay`. This would take more work to address; most people who currently import `lib/grpc.dart` will have a conflict between the protobuf Duration and the one from dart.core.

devoncarew:
cc @mosuem for review

Issue #736, opened 2024-09-19
lib/src/server/handlers.dart 's _customHeaders is reset to null, it should be reset to empty Map instead


isoos:
Error report:
https://www.reddit.com/r/dartlang/comments/1fk84bb/unhandled_exception_from_grpc_package/

The reported line seems to be when forcing `_customHeaders` to not-null: https://github.com/grpc/grpc-dart/blob/master/lib/src/server/handler.dart#L384

Looking at the git blame history, I think `_customHeaders = null` should have been migrated to `_customHeaders = <String, String>{}` when the null-safety migration happened.

mraleph:
I think the current code is correct. You can only send headers once so they are intentionally reset to `null` after that. You should not be able to enter this code again (it is additionally guarded by checking `_headersSent`). 

I also see that the [benchmark code in question](https://github.com/LesnyRumcajs/grpc_bench/blob/master/dart_grpc_onhold/Dockerfile) is referencing Dart 2.19 and `pubspec.lock` mentions grpc version 3.1. This is really old stuff. 

We will need to know exact stack trace and exact version of grpc package for this to be actionable. 

For now I am going to close this - but if [r/David_Owens](https://www.reddit.com/user/David_Owens/) provides necessary info I can reopen.

owensdj:
Here is a screen shot of the stack trace. It's running in a Docker container and only comes up for a split second, so I had to take a video of it.

https://1drv.ms/i/c/29ea63e9e7ad0202/EQ3bEw8Cer1KrEQHk-vpYmABnqFhUoNf85uwvdy7C0MLqQ?e=oGMD3A

mraleph:
@owensdj what is the version of the package / Dart SDK you are using?

owensdj:
gRPC 4.0.1 and Dart SDK 3.5.2. The problem happened with older versions as well.

mraleph:
One possible issue is that here https://github.com/grpc/grpc-dart/blob/master/lib/src/server/handler.dart#L360-L367 `_stream.sendHeaders` can trigger an error at which point `_customHeaders` will end up `null` but `_headersSent` is not `null`. This might be an underlying root cause.

owensdj:
[...]

PR #801, opened 2025-09-01
Fix null custom headers

hiro-at-pieces:


mraleph:
You need to sign CLA otherwise we can't even look your patch.

mosuem:
Is this to solve https://github.com/grpc/grpc-dart/issues/736? From the discussion there, I don't believe this is the right fix. Could you elaborate?

tsavo-at-pieces:
## Detailed Code Review

### Changes
This PR fixes a null pointer exception in `ServerHandler` by replacing `_customHeaders = null;` with `_customHeaders = <String, String>{};` in two locations:
1. In `sendHeaders()` (line 367) -- after headers are sent
2. In `sendTrailers()` (line 396) -- inside the `!_headersSent` block when headers haven't been sent yet

The intent is to prevent a `Null check operator used on a null value` crash if `_customHeaders` is accessed after headers have been sent.

### Code Quality

**The fix addresses a real bug**, but the approach has significant issues:

1. **Incomplete fix -- `_customTrailers` has the same problem**: The `_customTrailers` field is also set to `null` on line 403 (`_customTrailers = null;`) after trailers are sent. If `sendTrailers()` is called a second time (e.g., via `_sendError` after a response error), `_customTrailers!` on line 399 will throw the same NPE. This PR does not address that at all. If you are fixing null-safety for `_customHeaders`, `_customTrailers` should receive the same treatment.

2. **Semantic concern with the approach**: Replacing `null` with an empty map changes the semantics. Currently, `null` signals "headers have already been consumed/sent, don't touch them again." Replacing with an empty map means `headers` getter will return an empty map instead of `null`, which could mislead service code into thinking it can still add custom headers after they have been sent (the headers would silently be ignored). A better approach might be to:
   - Add null checks before the `!` operator calls, OR
   - Guard `sendTrailers` to return early if already called, OR
   - Keep the `null` assignment but add `if (_customHeaders == null) return;` or similar guard at the top of `sendTrailers`

3. **No PR description**: The PR body is empty -- there is no explanation of what bug this fixes, under what conditions the null crash occurs, or any reproduction steps. This makes it difficult for reviewers to evaluate whether the fix is correct and complete.

4. **Missing CHANGELOG entry**: The PR health bot flagged this. A CHANGELOG entry should be added.

### Tests
No tests are included. A test should demonstrate:
- The scenario that causes the NPE (e.g., calling `_sendError` after `sendHeaders` has already been called, or double-calling `sendTrailers`)
- That the fix prevents the crash

### Verdict
**NEEDS_WORK** -- The fix is partially correct but incomplete:
1. `_customTrailers` has the exact same null vulnerability and is not addressed
2. The empty-map replacement changes semantics in a potentially misleading way -- a null guard or early-return approach would be safer
3. A PR description explaining the crash scenario is needed
4. Tests and CHANGELOG entry are missing
5. Consider whether the real root cause is that `sendTrailers`/`_sendError` can be called multiple times, and guard against that instead

PR #811, opened 2025-11-12
Regenerate protos with protoc_plugin-24.0.0

osa1:
- protocolbuffers/protobuf commit 2d5d446cc (11/11/2025)
- gooleapis/googleapis commit cf0434f4b (11/11/2025)

PR #812, opened 2025-11-13
Update protos

mosuem:
Upgrading protos with new `googleapis` and `protobuf` versions.

Recreation of #811 .


mosuem:
I am wondering why the protos are exposed at all. I think they shouldn't be. I will try to remove the export, which is breaking in itself - I think I added them a while back for some g3 error. Let's see.

PR #816, opened 2025-12-05
Upgrade protobuf to 6.0.0 and add proto export

### LINKED ISSUE #814: Upgrade grpc-dart to protobuf 6.0.0
I tried using grpc+protobuf 6.0.0 to get well-known types support, but the resolution fails because there is no version of grpc that supports protobuf 6.0.0.

Please upgrade :)

### LINKED ISSUE #815: 5.0.0 removed previously exported ErrorInfo
Prior to 5.0.0, `ErrorInfo` was exported but it was quietly removed in 774fd15cee42391e7001a812c8936ee1482a3a82:

Image

Previously this allowed us to do checks like:
```dart
...
} on GrpcError catch (e) {
    if (e.details != null) {
      for (final detail in e.details!) {
        if (detail is ErrorInfo) {
          print(detail.reason);
          print(detail.domain);
        }

        // Some other detail type
        print(e);
      }
    }
}
```

But now that `ErrorInfo` is not exported, even if I compile `error_details.proto` myself, the runtime check `is ErrorInfo` fails because the `ErrorInfo` used internally in the grpc library is different from the one I compiled myself.

The best workaround I can think of is to do this every time,
```dart
      for (final detail in e.details!) {
        final casted = ErrorInfo.fromBuffer(detail.writeToBuffer());
      }
```

but this is inconvenient and I don't know at runtime if that GeneratedMessage is actually an ErrorInfo or not. Is there a recommended workaround or was this removed in error?

mosuem:
Fixes #814
Fixes #815

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 5.1.0
 
-- Upgrade to package `protobuf:6.0.0`.
+- Added `protos.dart` library.
+- Require `protobuf:6.0.0`.
 
 ## 5.0.0
 


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.1.0
+
+- Upgrade to package `protobuf:6.0.0`.
```
kevmoo:
```suggestion
- Added `protos.dart` library.
- Require `protobuf:6.0.0`.
```

kevmoo (approved):

PR #820, opened 2026-01-01
Skip handling when handler is closed to prevent processing on canceled calls

### LINKED ISSUE #67: Helloworld server crash (and is very slow)
Run server from hello world example.
Run client and kill it. Repeat many time
Time to time, killing the client cause server crash. See trace bellow.

Also  the client is very slow (4 request per sec). Don't know if related or not.

GRPC Version : from the repository (../..) (0.4.0 ?)

## Repro steps
See above.

## Details

Server Stack Trace:
```
Unhandled exception:
Bad state: Cannot add event after closing
#0      _StreamController.add (dart:async/stream_controller.dart:584)
#1      _StreamSinkWrapper.add (dart:async/stream_controller.dart:858)
#2      TransportStream.sendHeaders (package:http2/transport.dart:241:10)
#3      ServerHandler.sendTrailers (package:grpc/src/server/handler.dart:310:13)
#4      ServerHandler._sendError (package:grpc/src/server/handler.dart:359:5)
#5      ServerHandler._onResponse (package:grpc/src/server/handler.dart:248:7)
#6      _RootZone.runUnaryGuarded (dart:async/zone.dart:1316)
```

My Quick and dirty workaround : handler.dart: 248
      try {
        _sendError(grpcError);
      } catch(err) {
        print(err);
      }

HyprUsr:
## Summary
This PR fixes a critical server-side crash occurring when a gRPC client disconnects immediately after initiating a request. This issue typically results in an unhandled `Bad state: Cannot add event after closing` error, which brings down the entire server process.

## Related Issues
Fixes #67 and #558 . This behavior has been reported previously but remained unresolved until now.

## The Problem
When a client terminates abruptly (e.g., a process crash or network drop) immediately after sending a request, the server-side stream may transition to a closed state while the handler is still attempting to send data or events back. Attempting to interact with the closed stream without a state check throws a fatal "Bad state" exception.

As a matter of principle, a client-side disconnection should never be able to compromise the stability of the server process.

## The Solution
The fix introduces a safety check on the isCanceled flag prior to any interaction with the client stream. This ensures that if the connection is already dead, the server gracefully handles the state rather than attempting to write to a closed sink.

## How to Reproduce

1. Start the gRPC server.
2. Run the following Go script. It opens a stream, sends a single message, and then calls os.Exit(0) to simulate an abrupt termination without a clean shutdown.
3. Observe the server logs for the Bad state: Cannot add event after closing crash.

```GO
package main

import (
	"context"
	"log"
	"os"
	"time"

	pb "your/module/path/protos" 
	"google.golang.org/grpc"
)

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
	defer cancel()

	conn, err := grpc.DialContext(ctx, "localhost:50051",
		grpc.WithInsecure(),
		grpc.WithBlock(),
		grpc.WithDefaultCallOptions(grpc.WaitForReady(true)),
	)
	if err != nil {
		log.Fatalf("dial: %v", err)
	}

	client := pb.NewMyServiceClient(conn)
	stream, err := client.Chat(ctx) 
	if err != nil {
		log.Fatalf("open stream: %v", err)
	}

	if err := stream.Send(&pb.ChatRequest{Message: "hi"}); err != nil {
		log.Fatalf("send: %v", err)
	}

	// Terminate immediately to mimic abrupt client death.
	os.Exit(0)
}
```

### DIFF
--- a/lib/src/server/handler.dart
+++ b/lib/src/server/handler.dart
@@ -151,6 +151,7 @@ class ServerHandler extends ServiceCall {
   // -- Idle state, incoming data --
 
   void _onDataIdle(GrpcMessage headerMessage) async {
+    if (isCanceled) return;
     onDataReceived?.add(null);
     if (headerMessage is! GrpcMetadata) {
       _sendError(GrpcError.unimplemented('Expected header frame'));
@@ -276,6 +277,7 @@ class ServerHandler extends ServiceCall {
   // -- Active state, incoming data --
 
   void _onDataActive(GrpcMessage message) {
+    if (isCanceled) return;
     if (message is! GrpcData) {
       final error = GrpcError.unimplemented('Expected request');
       _sendError(error);
@@ -316,6 +318,7 @@ class ServerHandler extends ServiceCall {
   // -- Active state, outgoing response data --
 
   void _onResponse(dynamic response) {
+    if (isCanceled) return;
     try {
       final bytes = _descriptor.serialize(response);
       if (!_headersSent) {
@@ -336,10 +339,12 @@ class ServerHandler extends ServiceCall {
   }
 
   void _onResponseDone() {
+    if (isCanceled) return;
     sendTrailers();
   }
 
   void _onResponseError(Object error, StackTrace trace) {
+    if (isCanceled) return;
     if (error is GrpcError) {
       _sendError(error, trace);
     } else {
@@ -443,6 +448,7 @@ class ServerHandler extends ServiceCall {
   }
 
   void _onDoneExpected() {
+    if (isCanceled) return;
     if (!(_hasReceivedRequest || _descriptor.streamingRequest)) {
       final error = GrpcError.unimplemented('No request received');
       _sendError(error);


tsavo-at-pieces:
## Security & Compatibility Review

*Reviewed from the perspective of [open-runtime/grpc-dart](https://github.com/open-runtime/grpc-dart) fork maintainers.*

### Risk Assessment
- **Security risk:** Low (defensive improvement)
- **Compatibility risk:** Low
- **Upstream divergence:** Minimal

### Details

This PR adds `if (isCanceled) return;` guards to six methods in `handler.dart`:

1. `_onDataIdle()` -- prevents processing incoming data on a canceled call
2. `_onDataActive()` -- prevents processing active data on a canceled call
3. `_onResponse()` -- prevents sending responses on a canceled call
4. `_onResponseDone()` -- prevents sending trailers on a canceled call
5. `_onResponseError()` -- prevents error handling on a canceled call
6. `_onDoneExpected()` -- prevents done processing on a canceled call

**Security perspective:** This is purely defensive -- early-return on canceled streams prevents potential "Bad state: Cannot add event after closing" crashes. A client cannot weaponize this to bypass server logic, since the guard only skips work on *already-terminated* calls.

**Compatibility perspective:** No public API changes. Behavior change is strictly internal: methods that previously would crash on closed streams now gracefully no-op.

**Fork relevance:** Our fork (`open-runtime/grpc-dart`) already has overlapping race condition fixes in `_onResponse()`, `sendTrailers()`, and `_onDoneExpected()` using try-catch guards around stream operations. This PR takes a complementary approach (early-return vs. try-catch). Our fork's approach is more granular (we catch specific stream-closed errors and log them), while this PR is a broader, simpler guard. Both approaches are valid. If we cherry-pick this, we should verify it composes well with our existing try-catch guards rather than creating redundant protection.

**Concern:** The `_onResponse` guard may silently swallow legitimate responses if `isCanceled` is set prematurely due to a race condition. However, since `isCanceled` is only set by explicit cancel/timeout/error paths, this risk is low in practice.

### Recommendation
**MERGE** -- Sound defensive fix that prevents server crashes from abrupt client disconnects. Addresses longstanding issues #67 and #558. Our fork should evaluate cherry-picking this as a cleaner complement to our existing try-catch guards.

tsavo-at-pieces:
## Fork Maintainer Review (open-runtime/grpc-dart)

### Summary
Adds `if (isCanceled) return;` guard checks to six methods in `handler.dart`: `_onDataIdle`, `_onDataActive`, `_onResponse`, `_onResponseDone`, `_onResponseError`, and `_onDoneExpected`. This prevents the server from crashing with "Bad state: Cannot add event after closing" when a client disconnects abruptly mid-request.

### Assessment
- **Safe to merge into our fork?** Yes, with notes
- **Breaking changes?** No
- **Key concerns:**
  1. **We already have overlapping fixes in our fork.** Our fork (open-runtime/grpc-dart) already addresses the same class of crashes with a different strategy: we wrap risky operations in `try-catch` blocks and use `!_requests!.isClosed` guards in `_onResponse`, `sendTrailers`, and `_onDoneExpected`. See our `handler.dart` for details.
  2. **This PR's approach is complementary, not conflicting.** The `isCanceled` early-return is a "gate at the door" approach, while our try-catch is a "safety net inside the room" approach. Both are valid and can coexist -- belt-and-suspenders is actually ideal for production gRPC servers.
  3. **The PR is well-motivated.** The reproduction steps are clear, the problem statement is accurate, and the fix is minimal and targeted. The referenced issues (#67, #558) are real and long-standing.
  4. **No tests included.** This is understandable given the difficulty of reliably testing race conditions, but a test that simulates client disconnection during handler processing would strengthen the PR.

### Recommendation
**MERGE** -- This is a clean, minimal, and correct fix for a real server stability issue. If this lands in upstream, we should cherry-pick it into our fork alongside our existing try-catch guards. The two approaches complement each other: `isCanceled` prevents entering methods at all, while our try-catch handles the edge cases where `isCanceled` may not yet be true but the stream is already closing. Having both provides defense-in-depth.

We have been running equivalent protections in production (open-runtime/grpc-dart) and can confirm this class of fix is critical for server reliability.

